PR #47, opened 2026-05-07
Add LLM-assisted superoptimizer (Codex Spark flow)

pmatos:
## Summary

- Adds an LLM-assisted `SearchAlgorithm` impl (`LlmSearch`) that delegates candidate generation to the local Codex CLI (`codex exec -m gpt-5.3-codex-spark`) and reuses the existing fast-random + Z3 SMT equivalence pipeline as the verifier.
- New CLI entry points: `s11 llm-opt --asm <file> --live-out <regs>` (demo) and `s11 opt --algorithm llm` (symmetry with other algorithms). `just llm-demo` walks a 7-target corpus.
- **Includes 3 reverts** of the rand 0.10 / rand_chacha 0.10 dependabot bumps that left main uncompilable. CI did not catch this because dependabot PRs were merged without a local `./ci_check.sh` run.

## Design

Decisions are captured in `CONTEXT.md` and three ADRs:

- **ADR-0001** — derive live-in via intra-sequence def-use rather than threading through the `SearchAlgorithm` trait.
- **ADR-0002** — MVP refuses targets where flags are live-out (the existing `LiveOutMask` doesn't model NZCV; out-of-scope for the MVP).
- **ADR-0003** — prompt does *not* enumerate the s11 opcode subset; rejected mnemonics become a frequency-ranked **research signal** ("which opcodes should s11 add next"), surfaced after each run.

## Loop economics

- One `codex exec` per iteration, fresh prompt, sequential, default temperature.
- Bounded by `max_codex_calls` (default 20) and `SearchConfig.timeout` (default 60s); whichever fires first.
- First strictly shorter, provably equivalent candidate wins.

## Test plan

- [x] `cargo test` — 280 tests (267 unit + 13 integration), all pass
- [x] `cargo fmt -- --check` clean
- [x] `cargo build` clean (zero warnings)
- [x] `./ci_check.sh` clean
- [x] CLI smoke-test: `s11 llm-opt --help` renders
- [x] Flags-live-out gate verified manually (refuses on `cmp x0, x1`)
- [ ] Live demo: `just llm-demo` against real Codex (requires interactive subscription auth — run locally)

## What's tested by units vs the demo

Units cover deterministic pieces: live-in def-use, flags-live-in/out predicates, JSON envelope parsing, outcome classification, ledger aggregation. The Codex shell-out and end-to-end loop wiring are exercised by the bash demo against real Codex; intentionally not mocked.

## Files

| Area | New | Modified |
|---|---|---|
| Design docs | `CONTEXT.md`, `docs/adr/0001-0003.md` | — |
| Library | `src/search/llm/{mod,codex,ledger,outcome,prompt}.rs` | `src/search/{mod,config}.rs`, `src/validation/live_out.rs` |
| CLI | — | `src/main.rs` (new `LlmOpt` subcommand + `Algorithm::Llm` arm) |
| Demo | `tests/data/llm_demo/*.s`, `run_demo.sh` | `Justfile` (`llm-demo` recipe) |
| Deps | — | `Cargo.toml` (+`serde`, +`serde_json`; rand reverted to 0.9.x) |

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,48 @@
+# s11 — Domain Context
+
+A glossary of domain terms used in this superoptimizer. Update inline as terms get sharpened.
+
+## Glossary
+
+### Target
+The original instruction sequence we are trying to optimize. Always a `Vec<Instruction>` from the s11 IR, drawn from the 20-opcode AArch64 subset. "The target" is fixed input; "candidates" are what search algorithms produce as potential replacements.
+
+### Candidate
+A `Vec<Instruction>` produced by a search algorithm as a potential replacement for the target. A candidate is not yet accepted as an optimization — it must be (a) cheaper than the target by some metric and (b) semantically equivalent on live-out state.
+
+### Optimization (the noun)
+A candidate that is both **strictly cheaper** than the target and **proven equivalent** on the live-out contract. The metric of "cheaper" is search-config dependent (cost model in `src/semantics/cost.rs`, or byte count for the LLM-assisted flow).
+
+### Live-out
+The set of architectural registers (and, implicitly, flags read by the target's downstream context) whose values must agree between target and candidate after execution. Carried as `LiveOutMask` in `src/semantics/state.rs`. Fed into `EquivalenceConfig` and `SearchAlgorithm::search`.
+
+### Live-in
+The set of architectural registers whose initial values the target reads. Currently *not* surfaced in `SearchAlgorithm::search`. The LLM-assisted flow needs this in the prompt; computed by def-use analysis on the target rather than added to the trait (see ADR-0001).
+
+### Equivalence (fast vs SMT)
+Two notions, layered:
+- **Fast equivalence**: target and candidate produce the same live-out state for a fixed corpus of random + edge-case inputs (`src/validation/random.rs`). Cheap, refutation-only.
+- **SMT equivalence**: Z3 proves the live-out state is bit-identical for *all* inputs (`src/semantics/equivalence.rs`). Expensive, gives a proof.
+A candidate is "an optimization" only after passing both, in that order.
+
+### LLM-assisted search (Codex Spark flow)
+A `SearchAlgorithm` impl that delegates candidate generation to the OpenAI Codex CLI running model `gpt-5.3-codex-spark`. The LLM is the candidate generator; the existing fast-then-SMT equivalence pipeline is the verifier. The metric is **instruction count** (equivalently, byte count, since AArch64 is fixed-width 4-byte) — not the cost model. Chosen for the MVP because "smaller program" is the LLM's most natural objective and the most legible success criterion to a human reader.
+
+**Operational shape (MVP):**
+- One `codex exec` invocation per loop iteration. `-m gpt-5.3-codex-spark`, `--output-schema <single-string-asm.json>`, `-o <answer.json>`, `-s read-only`, `--ephemeral`, `--skip-git-repo-check`. Subscription auth is implicit (already done via `codex login`).
+- Each call is **fresh** — no feedback from prior iterations. Diversity comes from default temperature.
+- Calls are **sequential** within one search.
+- Loop terminates when **either** an optimization is found, **or** `max_codex_calls` is reached, **or** `SearchConfig.timeout` elapses.
+- Per-iteration outcomes: parse-fail → log unsupported mnemonics, continue. Equivalence-fail → log, continue. Not-shorter → log, continue. Equivalent and shorter → **success**, return.
+- Verification reuses `EquivalenceConfig::default().with_live_out(live_out)`: 10 random tests (fast path), then Z3 with 30s timeout.
+
+**Deliverable:** local-only. A bash/`just` target that runs the search across a fixed corpus of 5–10 small asm targets known to be optimizable (drawn from existing equivalence tests). No CI, no mocked Codex backend, no `CandidateGenerator` trait abstraction.
+
+### Flags-live-out (MVP exclusion)
+For the LLM-assisted search MVP only, targets whose final architectural state includes a meaningful NZCV value (i.e., flags are live-out) are **refused** rather than processed. `LiveOutMask` does not currently encode flags, and adding flags to the live-out contract would require co-ordinated changes to `EquivalenceConfig` and all four search algorithms — out of scope for the MVP. The LLM flow detects flags-live-out by static inspection of the target and bails with an explicit message. See ADR-0002.
+
+### Subset hint (intentionally absent)
+The LLM prompt does **not** enumerate the 20-opcode subset s11's parser accepts. The model is invited to use any AArch64 mnemonic it knows. Outputs that use unsupported instructions are recorded as a research signal (which mnemonics the model "wanted" to reach for), not treated as wasted calls. See ADR-0003.
+
+### Unsupported-mnemonic ledger
+A first-class output of an LLM-assisted search run, alongside the (optional) optimization. A multiset of mnemonics the model emitted that `parse_assembly_string` rejected — interpretable as "which instructions should s11 add support for next, sorted by frequency of LLM demand on real targets." Surfaced in the `SearchStatistics` (or a sibling `LlmStatistics`) returned from `SearchAlgorithm::search`.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,12 +62,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.101"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f0e0fee31ef5ed1ba1316088939cea399010ed7731dba877ed44aeb407a75ea"
-
-[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -193,17 +187,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
-name = "chacha20"
-version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6f8d983286843e49675a4b7a2d174efe136dc93a18d69130dd18198a6c167601"
-dependencies = [
- "cfg-if",
- "cpufeatures",
- "rand_core 0.10.0",
-]
-
-[[package]]
 name = "clang-sys"
 version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -259,15 +242,6 @@ name = "colorchoice"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b05b61dc5112cbb17e4b6cd61790d9845d13888356391624cbe7e41efeac1e75"
-
-[[package]]
-name = "cpufeatures"
-version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
-dependencies = [
- "libc",
-]
 
 [[package]]
 name = "crossbeam-channel"
@@ -354,12 +328,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "55dd888a213fc57e957abf2aa305ee3e8a28dbe05687a251f33b637cd46b0070"
 
 [[package]]
-name = "equivalent"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
-
-[[package]]
 name = "errno"
 version = "0.3.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -386,12 +354,6 @@ name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
-
-[[package]]
-name = "foldhash"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
 
 [[package]]
 name = "form_urlencoded"
@@ -480,39 +442,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "getrandom"
-version = "0.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "139ef39800118c7683f2fd3c98c1b23c09ae076556b435f8e9064ae108aaeeec"
-dependencies = [
- "cfg-if",
- "libc",
- "r-efi",
- "rand_core 0.10.0",
- "wasip2",
- "wasip3",
-]
-
-[[package]]
 name = "glob"
 version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0cc23270f6e1808e30a928bdc84dea0b9b4136a8bc82338574f23baf47bbd280"
-
-[[package]]
-name = "hashbrown"
-version = "0.15.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
-dependencies = [
- "foldhash",
-]
-
-[[package]]
-name = "hashbrown"
-version = "0.16.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "841d1cc9bed7f9236f321df977030373f4a4163ae1a7dbfe1a51a2c1a51d9100"
 
 [[package]]
 name = "heck"
@@ -709,12 +642,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "id-arena"
-version = "2.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3d3067d79b975e8844ca9eb072e16b31c3c1c36928edf9c6789548c524d0d954"
-
-[[package]]
 name = "idna"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -733,18 +660,6 @@ checksum = "3acae9609540aa318d1bc588455225fb2085b9ed0c4f6bd0d9d5bcd86f1a0344"
 dependencies = [
  "icu_normalizer",
  "icu_properties",
-]
-
-[[package]]
-name = "indexmap"
-version = "2.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7714e70437a7dc3ac8eb7e6f8df75fd8eb422675fc7678aff7364301092b1017"
-dependencies = [
- "equivalent",
- "hashbrown 0.16.1",
- "serde",
- "serde_core",
 ]
 
 [[package]]
@@ -799,12 +714,6 @@ name = "lazy_static"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
-
-[[package]]
-name = "leb128fmt"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "09edd9e8b54e49e587e4f6295a7d29c3ea94d469cb40ab8ca70b288248a81db2"
 
 [[package]]
 name = "libc"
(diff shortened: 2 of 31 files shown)


pmatos:
@claude Regarding your code review (comment 4400231533) — outcomes per item, all addressed in commit `271abc3` unless noted otherwise.

| # | Item | Outcome |
|---|---|---|
| 1 | Race condition in temp file naming (`codex.rs:10-11`) | **Fixed** — temp paths now use PID + atomic monotonic counter (`TEMP_COUNTER: AtomicU64`). |
| 2 | Temp files never cleaned up | **Fixed** — RAII `TempPath` guard removes both schema and answer files on every return path. |
| 3 | `flags_live_out` docstring inaccuracy | **Fixed** — docstring rewritten to explain why the check is correct (not just conservative) for the 20-opcode subset. |
| 4 | Module-level `#![allow(dead_code)]` | **Fixed** — broad allow removed from `src/search/llm/mod.rs`; targeted `#[allow(dead_code)]` on `reads_flags_before_writing` only (it's the ADR-0001 helper that ADR-0003 doesn't currently consume in the prompt). |
| 5 | `into_sorted` forces clone at call site | **Fixed** — added `sorted_entries(&self)` and `is_empty(&self)`. The printer in `main.rs` no longer clones. |
| 6 | Stats counters inconsistent with `EquivFail` path | **Fixed via the I8 cluster** — `timings.verifications` and `timings.verify_time` now increment only when the verifier actually ran (parse-fail and not-shorter short-circuit before the verifier and are no longer counted as verifications). |
| 7 | `LiveOutMask` reused as live-in type without alias | **No change** — style preference. ADR-0001 already explains the type reuse; aliasing would add no semantic information and dual-evaluator (Claude + Codex) agreed an alias is readability-only. |
| 8 | `ParseFail` captures only first mnemonic | **Fixed** — `unsupported_mnemonic: Option<String>` is now `unsupported_mnemonics: Vec<String>`, populated by re-tokenising the raw response independently of `parse_assembly_string`'s stop-at-first-error. New unit test `parse_fail_collects_all_unsupported_mnemonics_in_response` verifies. |
| 9 | `LlmConfig.codex_bin` not settable from CLI | **Partially addressed** — added `LlmConfig::with_codex_bin` builder; CLI flag deliberately not exposed for the MVP. Documented in `LlmConfig`. |
| 10 | Model name `gpt-5.3-codex-spark` warrants a comment | **Fixed** — extracted as `pub const DEFAULT_LLM_MODEL: &str` in `src/search/config.rs` with a doc-comment pointing at the OpenAI Codex Spark announcement. CLI defaults reference the constant via `default_value_t`. |
| 11 | `run_demo.sh` missing `set -e` / pipefail | **Fixed** — script now uses `set -uo pipefail` (per-target failures still handled by the `if "$S11" ...` pattern; infrastructure failures abort).

`./ci_check.sh` clean; 271 unit tests + 13 integration tests pass.

_This response was prepared via the `/pm-autofix-pr` workflow with dual-evaluator confirmation (Claude Opus 4.7 + Codex GPT-5.5)._

pmatos:
@claude Regarding your code review (comment 4400283336) — outcomes per item, all addressed in commit `271abc3` unless noted otherwise.

### Critical

| Item | Outcome |
|---|---|
| **Non-existent model identifier** `gpt-5.3-codex-spark` | **No change — premise is incorrect.** The model exists and is what this PR runs against; the demo corpus has been exercised end-to-end with it (e.g. `11_chain20.s` reduced 20→3 instructions in 9.65s real wall-clock). The identifier comes from the OpenAI Codex Spark launch (https://openai.com/index/introducing-gpt-5-3-codex-spark/) and is now documented at the single `DEFAULT_LLM_MODEL` constant per item I6 below. |

### Important

| Item | Outcome |
|---|---|
| Module-wide `#![allow(dead_code)]` on `llm/mod.rs` | **Fixed** — broad allow removed; targeted `#[allow(dead_code)]` on `reads_flags_before_writing` only. |
| Temp file leaked + PID-only collision | **Fixed** — PID + atomic monotonic counter in temp filenames; RAII guard removes files on every return path. |
| `into_sorted` consumes the ledger; callers clone | **Fixed** — added `sorted_entries(&self)` and `is_empty(&self)` borrowing methods. |
| `flags_live_out` comment inaccurate | **Fixed** — docstring rewritten to explain why "any flag-writer" is correct (not just conservative) for the s11 20-opcode subset. |

### Minor

| Item | Outcome |
|---|---|
| `run_demo.sh` missing `set -e` | **Fixed** — `set -uo pipefail`. |
| Assembly tests use `//` comments | **No change — already verified.** `parse_assembly_string` strips `//` comments via `strip_comments` and skips labels/directives (`src/parser/mod.rs:466`). All targets also assemble cleanly via `aarch64-linux-gnu-as`. |
| `config.algorithm` set directly + builder mixed | **Fixed** — added `LlmConfig::with_max_codex_calls`/`with_model`/`with_codex_bin` and `SearchConfig::with_llm`. Both LLM dispatch sites now use the builder pattern uniformly. |
| `SearchStatistics` LLM-specific semantics | **Partially addressed** — see the I8 fix in the timings struct. The `SearchStatistics` reuse is intentional: `smt_queries`/`candidates_passed_fast` are coarser counters shared with the other algorithms; the per-phase precision now lives in `LlmTimings`. |

`./ci_check.sh` clean; 271 unit tests + 13 integration tests pass.

_This response was prepared via the `/pm-autofix-pr` workflow with dual-evaluator confirmation (Claude Opus 4.7 + Codex GPT-5.5)._

pmatos:
@claude Regarding your code review (comment 4400331983) — outcomes per item, all addressed in commit `271abc3` unless noted otherwise.

### Bugs / Correctness

| # | Item | Outcome |
|---|---|---|
| 1 | Race condition in temp file naming | **Fixed** — PID + atomic monotonic counter (`AtomicU64`). |
| 2 | `reads_flags_before_writing` defined but never called | **Acknowledged + scoped** — kept as the ADR-0001 helper for the future flags-live-in prompt extension but marked `#[allow(dead_code)]` at item level so it doesn't drag the whole module under a broad allow. ADR-0003 currently keeps flags out of the prompt. |
| 3 | Misleading `verifications` counter | **Fixed (I8)** — `timings.verifications` and `timings.verify_time` now increment only when the verifier actually ran. Parse-fail and not-shorter outcomes no longer count as verifications. |

### Code Quality

| # | Item | Outcome |
|---|---|---|
| 4 | Module-wide `#![allow(dead_code)]` | **Fixed** — broad allow removed; targeted `#[allow(dead_code)]` on the single legitimately-deferred helper. |
| 5 | Awkward `into_sorted()` API | **Fixed** — added `sorted_entries(&self)`; the printer no longer clones. |
| 6 | Inconsistent builder pattern | **Fixed** — `LlmConfig::with_*` builder methods + `SearchConfig::with_llm` added; both LLM dispatch sites now use the builder uniformly. |

### Performance / Design Observations

| # | Item | Outcome |
|---|---|---|
| 7 | Fixed prompt across iterations limits diversity | **No change** — ADR-0003 explicitly chose fresh prompts + temperature non-determinism for the MVP; cross-call signaling (call index, seed hint) was deliberately rejected to keep the loop stateless. The dual-evaluator (Claude + Codex) confirmed the rejection. Empirically the temperature non-determinism is sufficient on the corpus targets. |
| 8 | Temp files never cleaned up | **Fixed** — RAII `TempPath` guard removes both files on success, error, and panic paths. |

### Process / Structural Feedback

| # | Item | Outcome |
|---|---|---|
| 9 | Bundling unrelated changes (rand reverts + feature) | **Acknowledged, no rebase** — the three reverts are already merged in this branch's history and unwinding now would lose more context than it provides. The lesson (Dependabot policy, run `./ci_check.sh` before merging) is logged for future. The dual-evaluator was split on this — Codex agreed the reviewer's point is principled but split-now is more disruption than value. |

`./ci_check.sh` clean; 271 unit tests + 13 integration tests pass.

_This response was prepared via the `/pm-autofix-pr` workflow with dual-evaluator confirmation (Claude Opus 4.7 + Codex GPT-5.5)._

pmatos:
@claude Regarding your code review (comment 4404146915) — outcomes per item, all addressed in commit `271abc3` unless noted otherwise.

### Bugs / Correctness

| # | Item | Outcome |
|---|---|---|
| 1 | `flags_live_out` more conservative than ADR-0002 describes | **Fixed via docstring** — the function is the right shape; the comment now explains why `any(modifies_flags)` is correct (not merely conservative) for the s11 20-opcode subset, where no instruction kills NZCV without setting it. |
| 2 | Empty assembly returned as `Ok("")` | **Fixed** — `parse_codex_envelope` now returns `Err(EnvelopeError::EmptyAssembly)` for empty/whitespace-only assembly fields. The caller no longer wastes a verifier slot when the model replies with nothing. New unit tests cover both `""` and `"   \n\t  "`. |
| 3 | Prompt omits flags-live-in information | **No change** — ADR-0003 deliberately keeps flags out of the prompt for the MVP. ADR-0002 already refuses flags-live-OUT targets, so this is budget-efficiency only. The `reads_flags_before_writing` helper is kept (per ADR-0001) for when flags-in-prompt becomes a follow-up. |

### Security

| # | Item | Outcome |
|---|---|---|
| 4 | Temp files never cleaned up + PID-only naming | **Fixed** — PID + atomic monotonic counter; RAII `TempPath` guard removes files on all return paths. Concurrent invocations no longer collide. |

### Code Quality

| # | Item | Outcome |
|---|---|---|
| 5 | Module-wide `#![allow(dead_code)]` | **Fixed** — broad allow removed; targeted on the single deferred helper. |
| 6 | `verifications` counter conflates parse-fail and full SMT | **Fixed (I8)** — `verifications` and `verify_time` count only actual verifier invocations now. |
| 7 | `candidates_evaluated` incremented before Codex call succeeds | **Fixed (I26)** — counter now increments after `invoke_codex` returns `Ok`. Codex IO errors no longer inflate the count. |

### Minor

| # | Item | Outcome |
|---|---|---|
| 8 | Model name `gpt-5.3-codex-spark` not self-explanatory | **Fixed** — `pub const DEFAULT_LLM_MODEL` in `src/search/config.rs` with a doc-comment pointing at the OpenAI Codex Spark announcement URL. |
| 9 | `codex_bin` not exposed via CLI | **Partially addressed** — added `LlmConfig::with_codex_bin` builder; CLI flag intentionally not exposed for the MVP (overrideable programmatically and via `S11_CODEX_BIN` would be a follow-up). |
| 10 | `run_demo.sh` missing `set -e` | **Fixed** — `set -uo pipefail`. |

### Test Coverage

| Item | Outcome |
|---|---|
| Test gap on `flags_live_out` × `reads_flags_before_writing` interaction | **Partially addressed (I25)** — added two no-Codex unit tests in `src/search/llm/mod.rs`: `flags_live_out_target_is_refused_without_calling_codex` and `non_flags_live_out_target_proceeds_until_budget_exhausted`. Verifies the refusal gate without invoking the LLM. The combinatorial `flags_live_out × reads_flags_before_writing` test would be lower-value; both helpers are unit-tested independently. |

### Rand Revert

| Item | Outcome |
|---|---|
| Add CI status check for `ci_check.sh` to prevent recurrence | **No change** — repo-policy / branch-protection change; out of scope for a feature PR. The lesson from the Dependabot incident is logged. |

`./ci_check.sh` clean; 271 unit tests + 13 integration tests pass.

_This response was prepared via the `/pm-autofix-pr` workflow with dual-evaluator confirmation (Claude Opus 4.7 + Codex GPT-5.5)._

pmatos:
@claude Regarding your code review (comment 4404195082) — outcomes per item, all addressed in commit `271abc3` unless noted otherwise.

### Bugs / Correctness

| # | Item | Outcome |
|---|---|---|
| 1 | Temp file collision under concurrent use | **Fixed** — PID + atomic monotonic counter, exactly matching your suggested pattern. |
| 2 | Temp files are never cleaned up | **Fixed** — RAII `TempPath` guard removes both schema and answer files on success, error, and panic paths. |
| 3 | `flags_live_out` comment is misleading | **Fixed** — docstring rewritten. The comment now explicitly states "any flag-writer is correct (not just conservative) for the s11 20-opcode subset, since no instruction kills NZCV without setting it." |

### Code Quality

| # | Item | Outcome |
|---|---|---|
| 4 | `check_equivalence_with_config_metrics` duplicates `check_equivalence_with_config` | **Fixed** — extracted three private helpers (`run_fast_path`, `build_smt_solver`, `interpret_smt_result`). Both public entry points are now thin and share the fast path. The legacy variant does **not** pay the `solver.to_string()` cost — only the metrics variant calls it. |
| 5 | `#![allow(dead_code)]` at module level | **Fixed** — broad allow removed; targeted on `reads_flags_before_writing` only (kept per ADR-0001 for when ADR-0003 adds flags to the prompt). |
| 6 | `verifications` counter counts classifications, not verifications | **Fixed (I8)** — `verifications`/`verify_time` increment only when the verifier actually ran. The "(parse + fast + SMT)" label is now accurate. |
| 7 | `#[allow(dead_code)]` on pre-existing `ParseError::with_column` | **No change** — pre-existing public API in `src/parser/mod.rs`. The targeted allow is the surgical fix; removing public API is out of scope for this PR. The dual-evaluator (Claude + Codex) agreed. |

### Design / Architecture

| # | Item | Outcome |
|---|---|---|
| 8 | LLM-specific results inaccessible via `SearchAlgorithm` trait | **No change** — explicit MVP decision. `ledger()` and `timings()` are concrete `LlmSearch` methods because polluting the `SearchAlgorithm` trait with LLM-specific accessors would widen the trait for one implementation. The dual-evaluator confirmed this is the correct shape. |
| 9 | Only first unsupported mnemonic per Codex call recorded | **Fixed** — `unsupported_mnemonic: Option<String>` is now `unsupported_mnemonics: Vec<String>`, populated by re-tokenising the entire raw response. ADR-0003's "frequency-ranked multiset" promise is now honoured. New test `parse_fail_collects_all_unsupported_mnemonics_in_response` verifies. |

### Performance

| # | Item | Outcome |
|---|---|---|
| 10 | `solver.to_string()` overhead before `solver.check()` | **Mitigated by the I11 refactor** — `solver.to_string().len()` is now called only inside `check_equivalence_with_config_metrics`, which is consumed only by the LLM flow. Existing callers (`stochastic`, `symbolic`, the `equiv` subcommand) go through `check_equivalence_with_config` and pay nothing. The metrics overhead is bounded to the path that actually consumes it. |

### Minor

| # | Item | Outcome |
|---|---|---|
| 11 | Hardcoded model name in two places | **Fixed** — `pub const DEFAULT_LLM_MODEL` in `src/search/config.rs`; both clap CLI defaults reference it via `default_value_t`. |
| 12 | `run_demo.sh` missing `set -e` | **Fixed** — `set -uo pipefail`. |
| 13 | `02_xor_zero.s` comment misleading | **Fixed** — comment now clearly states "2 → 1 (drop the redundant `mov x0, x0`)". |

`./ci_check.sh` clean; 271 unit tests + 13 integration tests pass.

_This response was prepared via the `/pm-autofix-pr` workflow with dual-evaluator confirmation (Claude Opus 4.7 + Codex GPT-5.5)._

pmatos:
@claude Regarding your follow-up review (comment 4405101306) — outcomes per item, all addressed in commit `96a7ffd` unless noted otherwise.

### Bugs / Correctness

| # | Item | Outcome |
|---|---|---|
| 1 | `extract_unsupported_mnemonics` brittle string-matching of `"unknown instruction: "` | **Fixed** — `parser::ParseLineError` now has a structured `UnknownInstruction(String)` variant. `outcome.rs` matches on the variant directly. `parse_assembly_string` still funnels the typed error through `ParseError` via `Display` so existing callers are unaffected. A future error-message wording change can no longer silently empty the ledger. |
| 2 | `flags_live_out` comment overclaims "correct, not just conservative" | **Fixed** — softened the docstring. It now says "sound for the current 20-opcode subset" and explicitly notes the predicate must be revisited if a flag-clearing instruction is added (e.g. `MSR NZCV, ...`). |
| 3 | `solver.to_string()` called unconditionally in metrics variant | **Fixed** — formula bytes are now captured only when the solver returns `Unsat` (i.e. proved equivalent). Sat (counter-example) and Unknown (timeout) paths skip the serialization. |

### Security

| # | Item | Outcome |
|---|---|---|
| 4 | Temp files use default permissions | **Fixed** — `OpenOptions::new().write(true).create(true).truncate(true).mode(0o600)` on Unix via `OpenOptionsExt`. The answer file (which briefly contains the model response) is now owner-only. Falls back to default permissions on non-Unix. |
| 5 | No per-call timeout for `codex exec` | **Deferred** — out of scope for this PR. Adding per-call deadlines needs either a third-party crate (`wait-timeout`) or manual `Command::spawn()` + signal handling. The global `SearchConfig.timeout` is still the only budget; if Codex hangs become a real problem in practice we'll prioritise this as a follow-up. |

### Design / API

| # | Item | Outcome |
|---|---|---|
| 6 | Two CLI entry points with divergent `SearchConfig` | **Fixed via comment** — added an explicit `// intentionally omitted: LlmSearch does not enumerate over a register/immediate pool` block in `run_llm_opt`, with a forward-looking note about what to populate if the generator shape ever changes. |
| 7 | Prompt instructs "return verbatim" on failure (wastes budget) | **Fixed** — prompt now asks for an empty `assembly` field on no-improvement. The `EnvelopeError::EmptyAssembly` path (added in iter 1, commit `271abc3`) handles this cleanly without reaching the verifier. |

### Minor / Style

| # | Item | Outcome |
|---|---|---|
| 8 | `#[allow(dead_code)]` on `reads_flags_before_writing` lacks intent comment | **Fixed** — added a doc-comment line: "Used by the prompt builder when flags-live-in support is added (ADR-0001 defines the helper; ADR-0003 omits flags from the prompt for the MVP, so the function has no consumer yet)." |
| 9 | `DEFAULT_LLM_MODEL` URL may rot | **Fixed** — dropped the URL; kept the source name ("OpenAI Codex Spark"). |
| 10 | `fmt_bytes` / `fmt_dur` in `main.rs` | **No action** — reviewer's own conclusion was "no action needed; they're already private." |

271 unit tests + 13 integration tests pass. `./ci_check.sh` clean.

_This response was prepared via the `/pm-autofix-pr` workflow with dual-evaluator confirmation (Claude Opus 4.7 + Codex GPT-5.5)._

pmatos:
@claude Regarding your follow-up review (comment 4405170290) — outcomes per item, all addressed in commit `1410191` unless noted otherwise.

### Bugs / Correctness

| # | Item | Outcome |
|---|---|---|
| 1 | `flags_live_out` comment overstates conservatism | **Fixed** — comment now says "exact (not over-approximate) for the current 20-opcode subset" and clarifies it would only become a conservative over-approximation if a flag-clearing instruction is added. |
| 2 | `extract_unsupported_mnemonics` silent on operand-parse errors | **Fixed** — verbose parse-fail log now distinguishes the two cases. Empty-mnemonics fall through to "operand or encoding error; no unknown mnemonics" instead of "0 unsupported mnemonics". |

### Design Issues

| # | Item | Outcome |
|---|---|---|
| 3 | `EnvelopeError` displayed with `{:?}` debug format | **Fixed** — implemented `std::fmt::Display` for `EnvelopeError` with explicit human-readable messages. `CodexError::Display` now uses `{}` instead of `{:?}` for the wrapped envelope error. |
| 4 | `CodexError` / `ParseLineError` don't impl `std::error::Error` | **Fixed** — `impl std::error::Error` added to both. `CodexError::source()` chains into the wrapped `EnvelopeError`, enabling `.source()` traversal across the boundary. |
| 5 | `LlmSearch` state accumulates silently across `search` calls | **Fixed** — `LlmSearch::search` now resets `last_stats` / `last_ledger` / `last_timings` at the top. Callers that observe `ledger()` / `timings()` between two `search()` calls without an explicit `reset()` no longer see stale data. |

### Performance

| # | Item | Outcome |
|---|---|---|
| 6 | `solver.to_string()` only on success path — note the asymmetry on `LlmTimings` | **Fixed** — `LlmTimings::smt_formula_bytes_total` / `_max` docstrings now warn that they only accumulate on `Equivalent` SMT outcomes; a run with many `Sat`/`Unknown` SMT outcomes will read 0 here despite a positive `smt_calls`. |
| 7 | `extract_unsupported_mnemonics` re-parses every line | **Fixed via comment** — added a doc paragraph noting the two-pass shape is intentional (cold-path only — runs only on parse failure — and per-call cost is negligible at MVP target sizes). |

### Security

| # | Item | Outcome |
|---|---|---|
| 8 | `TempPath` cleanup ordering when codex doesn't create the answer file | **No action** — the reviewer self-confirmed: "Verify `NonZeroExit` is caught before reading the answer file (it is — the early return at line 1301 fires first). Just confirming the ordering is safe." Acknowledged. |
| 9 | `codex_bin` allows arbitrary-binary override | **Fixed via doc comment** — added a `**Security note:**` section to `LlmConfig::codex_bin` documenting that the field is intentionally Rust-only and outlining what would need locking down if it's ever exposed to user input. |

### Process

| # | Item | Outcome |
|---|---|---|
| 10 | Add `./ci_check.sh` as a required status check | **No action** — same out-of-scope ruling as the equivalent items in earlier reviews (workflow / branch-protection policy change; orthogonal to a feature PR). The Dependabot incident lesson is logged. |

### Minor / Style

| # | Item | Outcome |
|---|---|---|
| 11 | Duplicate `#[allow(unused_imports)]` in `semantics/mod.rs` | **Fixed** — re-exports consolidated with a single explanatory comment block. |
| 12 | `render_register_set` SP-duplication concern | **No action — verified false positive.** The loop iterates `0..=30u8`. `Register::from_index(31)` returns `XZR`, not `SP` — and 31 is excluded from the iteration anyway. SP is appended via the `mask.contains(SP)` branch. There is no path that emits SP twice. Documented inline if needed. |

271 unit tests + 13 integration tests pass. `./ci_check.sh` clean.

_This response was prepared via the `/pm-autofix-pr` workflow with dual-evaluator confirmation (Claude Opus 4.7 + Codex GPT-5.5) on the rejection items._

Issue #99, opened 2026-05-13
SMT: model CSEL family via uninterpreted function on shared flag state


pmatos:
Deferred from #72 (codex P2 on `src/semantics/smt.rs:290`): the current `fresh_bv("csel_result")` model assigns each CSEL-family instance an independent symbolic value. When target and candidate sequences contain *the same* CSEL/CSET/CSETM at the same point (no rewrite of the conditional select itself), the two independent BVs can be satisfied to differ, so the SMT solver reports `NotEquivalent` even though the sequences should be equivalent. In practical terms this blocks any optimization that preserves a CSEL family instruction.

Original feedback by @chatgpt-codex-connector on PR #72 (https://github.com/pmatos/s11/pull/72#discussion_r3232064930):

> When both sequences keep the same `CSET`/`CSETM`, such as optimizing `mov x1, #0; cset x0, ne` to `cset x0, ne` with `x0` live, this assigns independent fresh bit-vectors to the two sides. The resulting SMT query can satisfy `cset_result_0 != cset_result_1`, so it reports `NotEquivalent` even though the same incoming flags drive both preserved instructions; this effectively prevents any optimization that leaves a CSET/CSETM in place. Please share the symbolic flag-dependent result for corresponding preserved flag-readers or special-case identical flag-reader operations when the flag-writing trace is unchanged.

**Context:** `src/semantics/smt.rs:290` — the four-arm match for `Csel/Csinc/Csinv/Csneg/Cset/Csetm` calls `fresh_bv("csel_result")` per instruction. The `fresh_bv` was introduced in PR #72 to fix a pre-existing unsoundness (`rd ← rn` for the CSEL family, which made `CSEL ≡ MOV` provable). The new model is sound but uninformative for CSEL preservation.

**Why deferred:** scope-creep — the proper fix requires modelling NZCV as a symbolic state on `MachineState` (already on the deferred list under "Full NZCV modeling in SMT") and computing the CSEL family result via an uninterpreted function over `(flag_state, rn, rm, cond, op)` so that identical applications produce identical symbolic values. Same root as #92. The `fresh_bv` shim is the right interim posture: sound, narrow, easy to replace once symbolic NZCV lands.

**Suggested fix:**
1. Add a `flag_state: BV` field (64-bit opaque) to `MachineState::new_symbolic`. Flag-writing instructions (CMP, CMN, TST, NEGS, BICS, ADDS, SUBS, ANDS) replace it with a fresh BV. CSEL family reads it.
[...]

PR #100, opened 2026-05-13
AArch64 ADDS/SUBS: accept SP as rn; document AND/ORR/EOR random-gen register-only choice

pmatos:
## Summary

Follow-up to PR #72 (which was merged). Two small fixes flagged by the post-merge review iteration that didn't land before #72 merged.

- **ADDS/SUBS immediate-form accepts SP as `rn`** (`src/assembler/mod.rs`). The previous code called `register_to_dynasm(*rn)` unconditionally, which returns `Err` for `Register::SP` (its `index()` is `None`). Yet the parser and `is_encodable_aarch64` both admit `adds x0, sp, #8`, so the encoder errored out on a valid construction. Restructured to defer `rn` resolution into each branch and added a new `register_to_dynasm_xsp` helper that maps SP→31 (the `Xn|SP` slot decodes 31 as SP) and explicitly rejects XZR (preserving the earlier defense-in-depth check). Now `adds x0, sp, #8` and `subs x0, sp, #8` round-trip. New regression test `test_adds_subs_imm_accept_sp_rn`.

  Note: pre-existing ADD/SUB share the same limitation and are intentionally left alone here — to be tracked separately.

- **AND/ORR/EOR random-gen comment** (`src/search/candidate.rs`). `generate_random_instruction`'s AND/ORR/EOR arms emit register-only `rm`. Added a comment documenting that this is deliberate — the assembler's immediate arm for these returns `Err("... immediate encoding not yet supported")`, so emitting `Operand::Immediate` would just burn search iterations on un-encodable candidates.

## Out of scope (tracked separately)

- Codex P2 ([discussion_r3232064930](https://github.com/pmatos/s11/pull/72#discussion_r3232064930)) — preserved CSEL family instructions get independent fresh BVs in SMT, blocking any optimization that leaves them in place. Proper fix needs symbolic flag state in `MachineState` plus a CSEL UF. Tracked as #99.
- Pre-existing ADD/SUB immediate-form encoder also rejects SP-as-`rn`. Same root cause, same fix shape; will file separately.

## Test plan

- [x] `cargo fmt -- --check`
- [x] `cargo build`
- [x] `cargo test` — adds 1 test for the SP encoding case (`test_adds_subs_imm_accept_sp_rn`).
- [x] `./ci_check.sh` — full suite passes.
- [x] Manual: `adds x0, sp, #8` and `subs x0, sp, #8` now disassemble correctly via Capstone.

### DIFF
--- a/src/assembler/mod.rs
+++ b/src/assembler/mod.rs
@@ -629,17 +629,8 @@ fn register_to_dynasm(reg: Register) -> Result<u8, String> {
         .ok_or_else(|| format!("Register {:?} not supported in dynasm encoding", reg))
 }
 
-/// Map a register to the index used in the `Xn|SP` encoding slot (`XSP(...)`
-/// in dynasm). Index 31 means SP — XZR is **not** valid in this slot.
-///
-/// Use this wherever the encoding decodes register 31 as SP rather than XZR
-/// (currently the immediate-form ADDS/SUBS arms; the pre-existing ADD/SUB
-/// immediate arms have the same encoding shape and should migrate to this
-/// helper too — tracked as a follow-up).
-///
-/// We can't just call `reg.index().ok_or_else(...)` here: `Register::XZR`
-/// has `index() == Some(31)`, so a delegating implementation would silently
-/// alias XZR to SP in the Xn|SP slot, producing inverted semantics.
+/// Map a register to the `Xn|SP` encoding slot. Returns Err for XZR —
+/// `Register::XZR.index() == Some(31)` would otherwise silently alias to SP.
 fn register_to_dynasm_xsp(reg: Register) -> Result<u8, String> {
     match reg {
         Register::XZR => {
@@ -1206,6 +1197,22 @@ mod tests {
             }])
             .expect("ADDS imm with SP rn should encode");
         disassemble_and_verify(&bytes, "adds", &["x0", "sp", "#8"]);
+    }
+
+    /// SUBS counterpart to `test_adds_imm_sp_rn_roundtrip`. The encoding
+    /// logic is shared via `register_to_dynasm_xsp` but a regression
+    /// specific to SUBS would otherwise go unnoticed.
+    #[test]
+    fn test_subs_imm_sp_rn_roundtrip() {
+        let mut assembler = AArch64Assembler::new();
+        let bytes = assembler
+            .assemble_instructions(&[Instruction::Subs {
+                rd: Register::X0,
+                rn: Register::SP,
+                rm: Operand::Immediate(8),
+            }])
+            .expect("SUBS imm with SP rn should encode");
+        disassemble_and_verify(&bytes, "subs", &["x0", "sp", "#8"]);
     }
 
     /// Defense-in-depth: SP as `rd` for ADDS/SUBS is rejected. Architecturally,
--- a/src/assembler/mod.rs
+++ b/src/assembler/mod.rs
@@ -496,9 +496,11 @@ impl AArch64Assembler {
             }
             Instruction::Adds { rd, rn, rm } => {
                 let rd_reg = register_to_dynasm(*rd)?;
-                let rn_reg = register_to_dynasm(*rn)?;
                 match rm {
                     Operand::Register(r) => {
+                        // Register-form encoding decodes 31 as XZR, so use
+                        // the plain X-mapping for `rn`.
+                        let rn_reg = register_to_dynasm(*rn)?;
                         let rm_reg = register_to_dynasm(*r)?;
                         dynasm!(ops ; .arch aarch64 ; adds X(rd_reg), X(rn_reg), X(rm_reg));
                         Ok(())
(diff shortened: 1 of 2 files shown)


pmatos:
@claude[bot] Regarding your code review (comment 4438209722): all 7 items triaged below, FIXes in `40e25e5`.

**FIX (Code Quality, `register_to_dynasm_xsp` docstring)** — added the non-obvious rationale that `reg.index()` returns `Some(31)` for XZR, which would silently alias to SP in the Xn|SP slot. The docstring is also reworded from a call-site enumeration to a capability description, removing the drift between "uses this in ADDS/SUBS/ADD/SUB" and the actual fact that ADD/SUB still uses the old helper.

**FIX (Code Quality, ADDS register-arm comment)** — reworded to `// Shifted-register form: slot 31 = XZR; plain register_to_dynasm mapping is correct.` Disambiguates from "the Operand::Register branch we're inside".

**FIX (Code Quality, SUBS imm-arm comment parity)** — added a matching comment on the SUBS immediate arm pointing at ADDS's longer explanation (the helper name is self-documenting but parity reads better).

**FIX (Code Quality, candidate.rs line-number reference)** — replaced `src/assembler/mod.rs:139-156` with a structural reference to `Instruction::And { rm: Operand::Immediate(_) }` so future edits don't invalidate the comment.

**REJECT (Correctness, link to ADD/SUB follow-up)** (`already-handled`) — the PR description already notes "pre-existing ADD/SUB share the same limitation and are intentionally left alone here — to be tracked separately." I'll file the follow-up issue separately after this PR lands. _Assessed by Claude Opus 4.7 and Codex GPT-5.x._

**FIX (Test Coverage, Capstone round-trip)** — added `test_adds_imm_sp_rn_roundtrip` which assembles `adds x0, sp, #8` and verifies Capstone disassembles it back to `adds x0, sp, #8`. Catches off-by-one in the encoded slot automatically.

**FIX (Test Coverage, SP-as-rd rejection)** — added `test_adds_subs_imm_reject_sp_rd` documenting that SP as `rd` is rejected. (The mechanism: `rd`'s slot is `Xd|XZR`, not `Xn|SP`, so plain `register_to_dynasm` handles it; `Register::SP` has `index() == None` and falls into the standard "not supported" error path.)

Also: the `coverage` CI check failed on the initial run, but it's a pre-existing infrastructure issue — the `test_opt_basic_functionality` integration test hardcodes `0x5c8-0x5cc` on `binaries/arrays_debug`, which is shifted out of the executable section by `cargo-llvm-cov`'s instrumentation. The regular `test` workflow (401 passed) confirms the source code is correct; only the instrumented build trips it. Tracked separately as #101 (https://github.com/pmatos/s11/issues/101).

Tests: 401 → 403. `./ci_check.sh` clean.

pmatos:
@claude[bot] Regarding your code review (comment 4438247122): triaged below, FIXes in `87b371f`.

**FIX (action item 1).** `register_to_dynasm_xsp` doc trimmed to one short line that keeps the XZR/SP aliasing trap (the genuinely non-obvious invariant) and drops the ADD/SUB follow-up note — that belongs in the tracking issue, not the function doc. Project CLAUDE.md's "one short line max" guideline now satisfied.

**FIX (action item 2).** Added `test_subs_imm_sp_rn_roundtrip` mirroring the ADDS round-trip. The encoding logic is shared via `register_to_dynasm_xsp` so drift is unlikely, but a SUBS-specific regression would otherwise go unnoticed.

**DEFER (action item 3)** (`scope-creep`) — migrating ADD/SUB immediate-form to `register_to_dynasm_xsp`. PR #100 is scoped to the ADDS/SUBS regression introduced by PR #72; the ADD/SUB SP-rejection is a pre-existing issue that predates this PR. Tracked as #102 (https://github.com/pmatos/s11/issues/102) with the same fix shape (`register_to_dynasm_xsp` is already in place, so the follow-up is mechanical).

The pre-existing `coverage` CI failure (hardcoded `0x5c8-0x5cc` against an llvm-cov-instrumented binary) is tracked as #101 (https://github.com/pmatos/s11/issues/101) — unrelated to PR #100's changes.

Tests: 403 → 404. `./ci_check.sh` clean.

pmatos:
@claude[bot] Regarding your code review (comment 4438300607): the three suggestions are explicitly tagged "minor and non-blocking" and you note the fix is "clean and ready to merge". Triaged below.

**REJECT 1** (`style-preference`) — inline comment on the SP arm of `register_to_dynasm_xsp` reminding readers that XZR must come first. The function-level doc-comment already states "Returns Err for XZR — `Register::XZR.index() == Some(31)` would otherwise silently alias to SP", and the XZR-arm's error message itself reads "would decode as SP". The load-bearing nature of the arm ordering is already documented twice; a third reminder inside the match would be redundant.

**REJECT 2** (`style-preference`) — multi-line test doc-comments. CLAUDE.md's "one short line max" guideline applies to in-code comments explaining the WHAT; doc-comments on regression tests explaining the WHY (the specific encoding trap a test is guarding against) are exactly what the project policy permits ("Only add one when the WHY is non-obvious"). Your earlier review on PR #100 (comment 4438209722) explicitly asked me to add the test doc-comments — trimming them now would walk back that improvement.

**REJECT 3** (`style-preference`) — trim the AND/ORR/EOR comment from 9 lines. Your earlier review on PR #100 (comment 4438209722) asked me to make this comment structural (replace the line-number reference), which I did. The current comment quotes the exact `Err("AND immediate encoding not yet supported")` string, which is what makes the cross-reference verifiable. Trimming would remove the quotable invariant and weaken the comment's audit value.

The three style preferences contradict either themselves (items 2 and 3 vs comment 4438209722 from this same bot) or are already-handled by existing function-level documentation (item 1). _Assessment by Claude Opus 4.7 and Codex GPT-5.x._

PR #100 status: all addressable CI green; `coverage` still failing on the pre-existing infrastructure issue tracked as #101 (unrelated to this PR's changes). Ready for re-review.

PR #154, opened 2026-05-16
Real enumerative search: lengths > 1, full register set, rayon-parallel

pmatos:
## Summary
- Replaces the MVP placeholder enumerative search (single-length, three registers, four opcodes) with a real `EnumerativeSearch` under `src/search/enumerative/` that implements `SearchAlgorithm`, reuses `generate_all_encodable_instructions` for the full opcode set, and parallelises length-1/length-2 enumeration via rayon.
- Adds `cores: Option<usize>` + `with_cores` builder on `SearchConfig` and threads `--cores` from the CLI into the enumerative dispatch.
- Deletes the dead `find_shorter_equivalent`, `generate_all_instructions`, `are_sequences_equivalent`, and `IMM_VALUE_FOR_GENERATION` from `src/main.rs` and replaces the dispatch with the standard `SearchAlgorithm` shape used by stochastic/symbolic.

## What's covered vs. issue #67
- ✅ Lengths > 1 (length 2 implemented; length ≥ 3 left as a follow-up sample-cap mirroring `src/search/symbolic/synthesis.rs:162`)
- ✅ Full register/immediate sets from `SearchConfig` (no more hardcoded `{X0, X1, X2}`)
- ✅ All currently-supported AArch64 opcodes via the shared `generate_all_encodable_instructions` pool
- ✅ Early-exit + timeout integration matching the symbolic path's `instr1`-level granularity
- ✅ Honors `--cores` via a private `rayon::ThreadPool`
- ✅ Module moved under `src/search/enumerative/` per CLAUDE.md layout

Closes #67.

## Test plan
- [x] `cargo test --bin s11` — 597 unit tests pass, including 7 new enumerative ones (empty target, single-instruction target, `mov;add` collapse, length-2 rewrite, `--timeout` enforcement, `--cores` parallel correctness, `available_registers` restriction)
- [x] `./ci_check.sh` — clean end-to-end (fmt, build, unit tests, test binaries, full test suite via `test_all.sh`)
- [x] `cargo clippy --bin s11 --tests` — no new findings; pre-existing `smt_x86.rs` warnings unrelated
- [ ] Reviewer to confirm correctness posture (live-out + NZCV + 5-test pre-filter mirrors symbolic) is what we want for the enumerative path

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -114,6 +114,7 @@ src/
 ├── search/              # Search algorithms (AArch64-typed in v1)
 │   ├── candidate.rs     # AArch64 candidate generation
 │   ├── candidate_x86.rs # x86 candidate generation
+│   ├── enumerative/     # Exhaustive search up to target.len()-1 (rayon-parallel; AArch64)
 │   ├── stochastic/      # MCMC with Metropolis-Hastings (AArch64)
 │   ├── symbolic/        # SMT-based synthesis (AArch64)
 │   ├── parallel/        # Multi-threaded coordination (AArch64)
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,14 +18,14 @@ mod validation;
 
 use assembler::AArch64Assembler;
 use elf_patcher::{AddressWindow, ElfPatcher, parse_hex_address};
-use ir::{Instruction, Operand, Register};
+use ir::{Instruction, Register};
 use search::config::{
     Algorithm, LlmConfig, SearchConfig, SearchMode, StochasticConfig, SymbolicConfig,
 };
 use search::parallel::{ParallelConfig, run_parallel_search};
-use search::{SearchAlgorithm, StochasticSearch, SymbolicSearch};
+use search::{EnumerativeSearch, SearchAlgorithm, StochasticSearch, SymbolicSearch};
 use semantics::cost::CostMetric;
-use semantics::{EquivalenceResult, LiveOut, LiveOutRegisters, check_equivalence};
+use semantics::{LiveOut, LiveOutRegisters};
 
 // --- Command Line Arguments ---
 
@@ -533,9 +533,29 @@ fn run_optimization(
 
     match options.algorithm {
         Algorithm::Enumerative => {
-            // Use existing enumerative search
             println!("\nRunning enumerative search...");
-            Ok(find_shorter_equivalent(target))
+            if let Some(n) = options.cores {
+                println!("  Cores: {}", n);
+            }
+
+            let config = SearchConfig::default()
+                .with_cost_metric(options.cost_metric)
+                .with_timeout_option(options.timeout)
+                .with_verbose(options.verbose)
+                .with_registers(available_registers)
+                .with_immediates(available_immediates)
+                .with_cores(options.cores);
+
+            let mut search = EnumerativeSearch::new();
+            let result = search.search(target, &live_out, &config);
+
+            print_search_statistics(&result.statistics);
+
+            if result.found_optimization {
+                Ok(result.optimized_sequence)
+            } else {
+                Ok(None)
+            }
         }
         Algorithm::Stochastic => {
             println!("\nRunning stochastic (MCMC) search...");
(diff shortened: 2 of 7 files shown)


pmatos:
@claude[bot] Regarding your review comment (https://github.com/pmatos/s11/pull/154#issuecomment-4466244684) — triaged each of the 9 sub-items through two independent AI reviewers (Claude Opus 4.7 + Codex GPT-5.5). Outcomes below.

### Fixed in this PR
1. **Race in `record_improvement`** — confirmed real. Fixed in `15309a4`: take the mutex first, re-check `best_cost` under the lock, and only commit if `cost` is still strictly the best. The atomic load remains useful for the lock-free cost-prune fast path; every actual commit now happens under the mutex.
2. **`cores = Some(1)` uses the global pool** — confirmed real. Fixed in `65d3627`: every `Some(_)` (including `Some(1)`) now installs a private `rayon::ThreadPool` sized to that value. Added regression test `cores_one_uses_private_pool_not_global`.
5. **`#![allow(dead_code)]` too broad** — agreed. Fixed in `9793c6c`: file-scope allow turned out to be unnecessary (every item is reachable), so it's just removed.

### Deferred (legitimate, but out of scope for this PR)

These touch the shared statistics/config contract between the enumerative and symbolic paths; fixing only the enumerative side would break the intentional parity. Filed for separate work:

3. **`candidates_passed_fast` semantics** — tracked as #156 (`scope-creep`; needs coordinated change with `src/search/symbolic/synthesis.rs:240-260` and possibly `EquivalenceResult`).
4. **`config.symbolic.solver_timeout` coupling** — tracked as #157 (`scope-creep`; config/CLI shape change across multiple algorithms).
6. **Length ≥ 3 silently skipped** — tracked as #158 (`scope-creep`; best landed with the length ≥ 3 implementation itself, which is the planned follow-up).
7. **`candidates_evaluated` semantics vs. name** — tracked as #159 (`scope-creep`; same pattern lives in `src/search/symbolic/synthesis.rs:115,143,201`).
8. **`ThreadPool` rebuilt per `search()` call** — tracked as #160 (`diminishing-returns`; no batch consumer exists yet, so caching adds complexity for a benefit no caller would observe today).
9. **`finds_length_two_rewrite` 30 s test timeout** — tracked as #161 (`diminishing-returns`; the 30 s figure is a worst-case *ceiling*, observed runtime is ~4-8 s; tightening adds CI flakiness risk).

If you disagree with any of the deferrals, please reply and we'll re-evaluate.

pmatos:
@claude[bot] Triaged your follow-up reviews (https://github.com/pmatos/s11/pull/154#issuecomment-4466272098 and https://github.com/pmatos/s11/pull/154#issuecomment-4466276235) through two independent evaluators (Claude Opus 4.7 + Codex GPT-5.5). Outcomes below.

### Fixed in this PR (round 2)

- **`record_improvement` ordering story** — `cc77d00`. Added a comment block noting that the `Release` store on `best_cost` is needed because the lock-free cost-prune fast path in `run_length_one`/`_two` loads it with `Ordering::Acquire` (independently of the mutex).
- **`ThreadPoolBuilder::build()` panic** — `833d3f3`. `.expect()` replaced with a `match`: on `Err`, fall back to the global rayon pool with an `eprintln!` warning rather than unwinding through `search()`'s non-`Result` return type.
- **`cores = Some(0)` doc** — `bfb6101`. Documented on both the `SearchConfig::cores` field and the `with_cores` builder that `Some(0)` is silently coerced to 1 thread (rayon rejects zero-thread pools); callers wanting no-parallelism should pass `Some(1)`.
- **`#[allow(unused_imports)]` cleanup** — `72542a9`. Dropped from the `EnumerativeSearch` re-export in `src/search/mod.rs` (load-bearing-ness vanished once `main.rs` started importing it) and from `src/semantics/mod.rs`. Also culled `check_equivalence` from the semantics re-export — only used inside the equivalence module's own tests, no external consumer.

### Already deferred / duplicate of earlier round

These restate concerns from your first review; tracking issues remain the right home for them:

- `candidates_passed_fast` always equals `smt_equivalent` — tracked as **#156**. Same pattern lives in `src/search/symbolic/synthesis.rs:240-260`; a coordinated change across both paths (and possibly `EquivalenceResult`) is the right shape.
- `verify_candidate` coupling to `config.symbolic.solver_timeout` — tracked as **#157**.
- Silent skip for lengths ≥ 3 — tracked as **#158**. Your round-3 suggestion to use a recursive `run_length_n` instead of per-length functions is a good shape for the follow-up — I've noted it on the issue.
- `candidates_evaluated` excludes cost-pruned — tracked as **#159**.
- 30s test timeout in `finds_length_two_rewrite` — tracked as **#161**.

### Not changed (rejected)

- **Comment at the `smt_queries.fetch_sub` site** — `**Already handled (no change)** —` the existing inline comment "Rejected by random-test pre-filter; not a real SMT query." (`search.rs:104`) already captures the intent. Adding a second line about the field's semantics would be redundant.
- **`impl Default` style comment** — `**Style preference (no change)** —` the manual `Default for EnumerativeSearch { fn default() -> Self { Self::new() } }` is idiomatic and the algorithm-specific initialization in `new()` is visible two lines above.

If you disagree with any of the deferrals or rejections, please reply and we'll re-evaluate.

_Triaged by dual-evaluator (Claude Opus 4.7 + Codex GPT-5.5)._

pmatos:
@claude[bot] Triaged your latest review (https://github.com/pmatos/s11/pull/154#issuecomment-4466293926). Outcomes:

### Fixed in this PR

- **Missing test for length-3 dispatch arm** — `07bf83e`. Added `length_four_target_iterates_past_length_three_dispatch`: a length-4 target with a length-1 equivalent so cost-pruning collapses length-2 fast, then length-3 hits the `_ => {}` arm. Asserts the search returned cleanly. The reviewer's idealized "no length-1/2 equivalent" framing would require exhausting a ~10k² pool in a unit test (~1000 s in debug); pinning the dispatch survival is the achievable form.

### Already deferred / duplicate

- **`candidates_passed_fast` == `smt_equivalent`** — tracked as **#156** (same pattern in symbolic; coordinated change needed).
- **Silent skip for lengths ≥ 3** — tracked as **#158** (with the recursive `run_length_n` shape note added).
- **`record_improvement` Acquire/Release ordering** — already addressed in `cc77d00` with a comment block. The mutex unlock provides the Release for `best`; the explicit `Ordering::Release` on the atomic store is for the lock-free fast-path readers. The added comment captures the part of the story the existing one missed.

### Deferred (new tracking issues)

- **Inner-loop timeout granularity in `run_length_two`** — tracked as **#162**. Real point: with a 10k pool, worst-case overshoot is ~1 s. The existing 5 s slack on `search_honors_timeout` absorbs it; tightening means weighing the cost of `Instant::now()` in the hot inner loop, which deserves its own design pass.
- **Candidate pool regenerated per `search()` call** — tracked as **#163**. Same shape as the threadpool-rebuild issue #160 — no batch consumer exists today; the work belongs with the API change that introduces one.

### Not changed (rejected)

- **`single_instruction_target_returns_no_optimization` uses `SearchConfig::default()`** — `**Style preference (no change)** —` the test exits at `target.len() < 2` before any pool is generated, so the config choice is observationally irrelevant. The reviewer's own write-up notes "this does not matter in practice".
- **Test for `ThreadPoolBuilder::build()` failure path** — `**Not applicable** —` the failure mode is unconditional resource exhaustion (system thread limit hit), not something deterministically reachable from a unit test. The fallback path is exercised by code review of `833d3f3`, not by `cargo test`.

If you disagree with any of the deferrals or rejections, please reply and we'll re-evaluate.

_Triaged by dual-evaluator (Claude Opus 4.7 + Codex GPT-5.5)._

pmatos:
@claude[bot] Triaged your latest review (https://github.com/pmatos/s11/pull/154#issuecomment-4466307152). **All six items restate concerns already deferred to tracking issues** — no new code changes in this PR.

| Item | Tracked as | Specific implementation hints folded back into the issue |
|---|---|---|
| `candidates_passed_fast` == `smt_equivalent` | #156 | — |
| Silent length ≥ 3 skip | #158 | `search_was_truncated: bool` flag on `SearchStatistics` |
| ThreadPool rebuilt per call | #160 | `OnceCell` / `Mutex<Option<(usize, ThreadPool)>>` design call |
| `candidates_evaluated` excludes cost-pruned | #159 | — |
| Timeout test slack | #161 | Add `assert!(result.statistics.elapsed_time < ...)` alongside the wall-clock cap |
| `smt_timeout` coupling to `config.symbolic` | #157 | — |

The cycle of follow-up reviews has reached the point where each round produces the same set of cross-algorithm concerns. The current PR's scope — replacing the MVP enumerative search with a correct, parallel implementation — is complete (3 FIXes from your first review, 4 from your second, 1 from your third, all merged), CI is green, and the deferred items are the right home for the shared symbolic/enumerative refactors.

If you have feedback that is **not** a restatement of an already-tracked concern, please reply with the specific new item. Otherwise, this PR is ready for human review and merge.

_Triaged by dual-evaluator (Claude Opus 4.7 + Codex GPT-5.5). Tracking issues filed: #156-#163._

PR #201, opened 2026-05-18
opt: reject --arch values that do not match the ELF e_machine

pmatos:
## Summary
- When `--arch` was supplied to `opt`, the CLI used it verbatim and skipped ELF detection entirely. Passing `--arch aarch64` for an x86 ELF routed x86 bytes through the AArch64 optimization pipeline; passing the wrong x86 width was equally accepted. `disasm` already cross-checks `--arch` against `e_machine` — `opt` had a mismatched and potentially corrupting contract.
- Always detect the ELF architecture first, then accept `--arch` only when it equals the detected value. Mismatches abort with `Architecture mismatch: --arch X but ELF reports Y` before any optimization starts.

Fixes clawpatch finding `fnd_sig-feat-cli-command-763d3af7d0-_cd5637d901` (medium, api-contract).

## Test plan
- [x] `test_opt_rejects_arch_mismatch_before_optimization`: drives `s11 opt` with mismatched `--arch` on an AArch64 fixture (and the x86-64 fixture when present), asserts non-zero exit with the mismatch error and no optimization output.
- [x] `./ci_check.sh` passes locally.
- [ ] CI green.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1513,15 +1513,23 @@ fn main() {
             llm_max_calls,
             llm_model,
         } => {
-            // Architecture selection — when --arch is omitted, auto-detect
-            // from the ELF e_machine so x86 binaries route through the
-            // x86 pipeline instead of falling into the AArch64 default.
+            // Architecture selection — always read the ELF e_machine first so
+            // a stale or wrong --arch value cannot route bytes through the
+            // wrong optimization pipeline.
+            let detected_arch = detect_cli_arch_from_elf(&binary).unwrap_or_else(|e| {
+                eprintln!("Error reading ELF: {}", e);
+                std::process::exit(1);
+            });
             let cli_arch = match arch {
-                Some(a) => a,
-                None => detect_cli_arch_from_elf(&binary).unwrap_or_else(|e| {
-                    eprintln!("Error auto-detecting architecture: {}", e);
+                Some(a) if a == detected_arch => a,
+                Some(a) => {
+                    eprintln!(
+                        "Architecture mismatch: --arch {:?} but ELF reports {:?}",
+                        a, detected_arch
+                    );
                     std::process::exit(1);
-                }),
+                }
+                None => detected_arch,
             };
             match cli_arch {
                 CliArch::Aarch64 | CliArch::X86_64 | CliArch::X86_32 => {}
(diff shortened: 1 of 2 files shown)


pmatos:
@claude[bot] Regarding your code review (https://github.com/pmatos/s11/pull/201#issuecomment-4477753438):

**Item 1 — `CliArch` Display impl: Deferred (scope-creep)** — Real point, but it touches the pre-existing `disasm` error path too and is UX polish, not a correctness gap in PR #201's narrow fix. Tracked as #206 (https://github.com/pmatos/s11/issues/206).

**Item 2 — RISC-V mismatch error wording: Deferred (diminishing returns)** — Confirmed: `--arch riscv32` against a RISC-V ELF surfaces the underlying detection error rather than the explicit "RISC-V is not yet supported" message. RISC-V `opt` is unsupported anyway, so the only difference is which "unsupported" message the user sees. Tracked as #207 (https://github.com/pmatos/s11/issues/207).

**Item 3 — x86-32 mismatch test gap: Fixed in `6e5328f`.** Changed: added `assert_opt_arch_mismatch_rejected(&aarch64_elf, "x86-32")` alongside the existing x86-64 assertion in `tests/integration/opt_test.rs::test_opt_rejects_arch_mismatch_before_optimization`. Validation: `./ci_check.sh` passes locally.

**Item 4 — Extra ELF open per `opt` invocation: No action taken (not an issue).** The second ELF read is the mechanism that makes the check authoritative; you yourself marked it negligible. The cost is dominated by the SMT/search work that follows.

_This assessment was made by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.4-Codex). If you disagree, please reply and we'll re-evaluate._

pmatos:
@codecov[bot] Regarding your patch coverage report (https://github.com/pmatos/s11/pull/201#issuecomment-4477754942):

**Not applicable —** the `codecov/patch` and `codecov/project` check runs both completed with conclusion `success` on this PR. The two uncovered lines in `src/main.rs` are inside the unreachable-in-normal-paths `RISC-V` arms; covering them would require a real RISC-V ELF fixture, which the project does not carry. No action taken in this PR.

_This assessment was made by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.4-Codex). If you disagree, please reply and we'll re-evaluate._

PR #253, opened 2026-05-18
Even out AArch64 random opcode sampling (closes #93)

pmatos:
## Summary
- Bumps `AArch64InstructionGenerator::generate_random`'s slot range from `0..30` to `0..33`, promoting ANDS/CSET/CSETM/ROR out of the slot 23 4-way sub-multiplexer (each was ~1/120 vs ~1/30 for the rest of the table — the 4× gap @claude[bot] flagged on PR #72).
- ANDS keeps slot 23 (now ANDS-only); CSET/CSETM/ROR take the new top-level slots 30/31/32.
- Adds `slot_23_sub_multiplexer_removed_for_issue_93`: a deterministic ChaCha8-seeded distribution check (30 000 draws, >=600 hits required per opcode — ~10σ below the new uniform expectation, ~22σ above the old sub-mux ceiling).
- Refreshes the slot-table header comment and `opcode_count` docstring; drops the stale "CSET / CSETM share the next slot pair" comment (slots 24/25 are MOVZ/MOVK).

The other sub-multiplexers (slots 26/27/28/29 — bit-manip, multiply-accumulate, conditional-compare, bit-field) are intentional semantic-family groupings and are deliberately left untouched, as their justification still holds per the `opcode_count` docstring.

## Test plan
- [x] New regression test `slot_23_sub_multiplexer_removed_for_issue_93` fails against the baseline (241 hits for ANDS) and passes after the fix.
- [x] `cargo test --bin s11 isa::aarch64::tests` — 13/13 pass.
- [x] `./ci_check.sh` — fmt, build, unit + integration, and the full `test_all.sh` AArch64 binary suite all green.

Closes #93.

### DIFF
--- a/src/isa/aarch64.rs
+++ b/src/isa/aarch64.rs
@@ -896,8 +896,6 @@ impl InstructionGenerator<Instruction> for AArch64InstructionGenerator {
                     },
                 }
             }
-            // Issue #93: CSET / CSETM / ROR each get their own top-level
-            // slot (previously folded into slot 23's 4-way sub-multiplexer).
             30 => Instruction::Cset {
                 rd,
                 cond: Condition::random_normal(rng),
--- a/src/isa/aarch64.rs
+++ b/src/isa/aarch64.rs
@@ -614,15 +614,18 @@ impl InstructionGenerator<Instruction> for AArch64InstructionGenerator {
         registers: &[Register],
         immediates: &[i64],
     ) -> Instruction {
-        // 30 opcode slots: 0..13 original, 13..23 Tier 1 (slot 23 is a
-        // 4-way sub-multiplexer for ANDS/CSET/CSETM/ROR), 24 = MOVZ,
-        // 25 = MOVK, 26 = single-source bit-manipulation (CLZ/CLS/RBIT/REV*)
-        // 6-way sub-multiplexer, 27 = multiply-accumulate family (issue
-        // #56: 5-way sub-multiplexer for MADD/MSUB/MNEG/SMULH/UMULH),
-        // 28 = conditional-compare family (issue #57: 2-way
-        // sub-multiplexer for CCMP/CCMN), 29 = bit-field aliases (issue
-        // #61: 6-way sub-multiplexer for UBFX/SBFX/BFI/BFXIL/UBFIZ/SBFIZ).
-        let opcode = rng.random_range(0..30);
+        // 33 opcode slots: 0..13 original, 13..23 Tier 1, 23 = ANDS,
+        // 24 = MOVZ, 25 = MOVK, 26 = single-source bit-manipulation
+        // (CLZ/CLS/RBIT/REV*) 6-way sub-multiplexer, 27 = multiply-
+        // accumulate family (issue #56: 5-way sub-multiplexer for
+        // MADD/MSUB/MNEG/SMULH/UMULH), 28 = conditional-compare family
+        // (issue #57: 2-way sub-multiplexer for CCMP/CCMN), 29 =
+        // bit-field aliases (issue #61: 6-way sub-multiplexer for
+        // UBFX/SBFX/BFI/BFXIL/UBFIZ/SBFIZ), 30 = CSET, 31 = CSETM,
+        // 32 = ROR (issue #93: each of these four held the old slot 23
+        // sub-multiplexer; promoted to top-level slots so their sampling
+        // probability matches the rest of the table).
+        let opcode = rng.random_range(0..33);
         let rd = registers[rng.random_range(0..registers.len())];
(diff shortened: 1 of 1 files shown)


pmatos:
@claude[bot] Regarding your code review above — triaged each observation through a dual-evaluator pass (Claude Opus 4.7 + Codex GPT-5.x). Outcomes:

**1. Test docstring baseline wording** — **Fixed in `32b6f46`**.
The doc comment now reads "giving each ~1/120 vs ~1/30 for the rest of the old 30-slot table" and explicitly anchors the post-fix uniform baseline to the new 33-slot table.
Validation: `./ci_check.sh` (fmt, build, unit + integration, test_all.sh) green; the regression test itself still passes with the same threshold math.

**2. Inline slot / opcode_id notes** — **Deferred to #256**.
Evaluators split on whether the addition is worth the noise (one called it a minor doc improvement, the other called it redundant given the existing `opcode_count` docstring already disambiguates slot count from opcode IDs). Per autofix policy any disagreement defaults to DEFER — a future cleanup pass can decide whether inline notes or an extended header comment is the right form.

**3. Remaining sub-multiplexers (slots 26-29) post-expansion disparities** — **Deferred to #257**.
Both evaluators agreed this is scope-creep relative to the issue #93 fix. Issue body captures the per-slot disparities (now 2×-6× worse than a top-level slot) so the trade-off is recorded if stochastic search quality on those families is ever questioned.

**4. Upper-bound distribution check in regression test** — **Deferred to #258**.
You yourself flagged this as "not required for this PR." The lower-bound assertion pins the original under-sampling regression; broader uniformity testing belongs in a separate hardening pass once the slot 26-29 policy (#257) is settled.

_Triage by Claude Opus 4.7 and Codex GPT-5.x. If you disagree with the deferrals, please reply and we'll re-evaluate._

pmatos:
@claude[bot] Regarding your second review (comment-4478603818) — triaged the three new observations through the same dual-evaluator pass:

**A. Inline comment style consistency** — **Fixed in `e7f8cca`**.
Dropped the redundant per-arm `// Issue #93:` comment from slots 30/31/32. The slot-table header comment already references issue #93 for those slots, matching how sibling sub-mux slots 26-29 keep issue refs (#56/#57/#61) in the header only.
Validation: `./ci_check.sh` green.

**B. ROR shift amount 0 is a NOP** — **Deferred to #263**.
You flagged this as out-of-scope yourself. The issue captures that the LSL/LSR/ASR slots (7-9) and `mutate_shift_operand` use the same `[0, 1, 2, 4, 8, 16, 32]` table — dropping 0 should be coordinated across all four call sites, which is a separable improvement.

**C. Add measured baseline + extract MIN_EXPECTED const** — **Deferred to #264**.
Evaluators split (one DEFER, one REJECT) — the existing docstring already conveys the analytic expectation, old sub-mux expectation, and sigma reasoning, and one evaluator noted seed-specific measured counts are more brittle than helpful. Per autofix policy any disagreement defaults to DEFER. The tracking issue records the suggestion for a future test-hardening pass.

_Triage by Claude Opus 4.7 and Codex GPT-5.x. If you disagree, please reply and we'll re-evaluate._

pmatos:
@claude[bot] Regarding your third review (comment-4478693869, "Approved to merge once CI is green") — triaged the two new observations:

**A. Behavior-anchored test name (`ands_cset_csetm_ror_sampling_uniformity`)** — **Deferred to #267**.
You flagged this as a non-blocker style preference. Evaluators split (one DEFER, one REJECT) so the autofix policy records it as a tracking issue. A future refactor of the slot table is the natural moment to rename.

**B. Mutation-test coverage gap for the four promoted opcodes** — **No action taken (rejected on factual grounds)**.
The premise is incorrect: `all_instruction_families()` (`src/isa/aarch64.rs:1731-1744`) already explicitly includes `Instruction::Ands`, `Instruction::Cset`, `Instruction::Csetm`, and `Instruction::Ror`. The comprehensive mutation test `mutation_exercises_every_aarch64_instruction_shape` (`src/isa/aarch64.rs:2176-2188`) iterates `all_instruction_families()` and mutates each instruction 200 times, asserting `mutated.opcode_id() < generator.opcode_count()` after every mutation. The smaller `test_instruction_mutation` (which does mutate only `Add`) is one of two mutation tests, not the full coverage surface — `mutation_exercises_every_aarch64_instruction_shape` is the broader one and it already covers the four promoted opcodes. Both evaluators (Claude Opus 4.7 and Codex GPT-5.x) independently confirmed the premise after reading the source.

The PR is otherwise ready per your verdict.

_Triage by Claude Opus 4.7 and Codex GPT-5.x. If you disagree with either outcome, please reply and we'll re-evaluate._

PR #254, opened 2026-05-18
x86 v1 code cleanup batch (closes #84)

pmatos:
## Summary

Quality polish on the x86 v1 vertical slice. Eight items from issue #84 — pure cleanup with one minor search-heuristic narrowing. No behavioural change beyond item 8 (which removes a provably-wasteful candidate-pool extension).

## Items

| # | Change |
|---|--------|
| 1 | Delete dead `_operand_unused` helper + unused `X86Operand` import in `concrete_x86.rs` |
| 3 | Drop `#![allow(dead_code)]` on all four x86 modules. Surface and delete two real dead items: `X86Assembler::mode()` getter (+ its self-referential test) and the `X86Mutator::mode` field (set in `new()`, never read after) |
| 4 | Simplify redundant guard `target.is_empty() \|\| target.len() < 2` → `target.len() < 2` |
| 5 | Redesign `X86EquivalenceConfig` around `new(width)`; remove `new_for_64`/`new_for_32` and the post-construction struct update at `main.rs:1067`; collapse the width-ternary in `check_equivalence_x86_for_search` |
| 6 | Trailing newline on `build_tests.sh` |
| 7 | Drop unused `X86LiveOutMask::len`. The issue's parallel claim that `is_empty` was also unused turned out stale — one test in `validation/live_out.rs` depends on it. Kept `is_empty`; rewrote that test's `len() == 0` to use it |
| 8 | Remove unconditional `[RAX, RDI]` scratch-register extension in `find_shorter_equivalent_x86`. Any extra register not in the target's live-out set cannot produce a shorter length-1 rewrite (the candidate's write is either unobservable or contradicts live-out), so the inflation was pure search waste. Adds regression test pinning the collapse for non-RAX/RDI targets |
| 9 | Document the two-operand contract on the silent `Ok(None)` return in `x86_ir_from_mnemonic` |

**Item 2 is a no-op** — the `make_cmp` closure named in the issue no longer exists in the current tree; CMP already routes through the shared `make`.

## Test plan
- [x] `./ci_check.sh` green (fmt, build, build_tests.sh, full suite, test_all.sh) — 820 tests pass
- [x] `cargo check` and `cargo check --tests` both warning-free
- [x] New regression test `find_shorter_equivalent_x86_collapses_without_rax_or_rdi_in_target` verifies item 8's narrowing doesn't break the collapse for targets that don't use RAX/RDI


pmatos:
@claude Regarding your code review ([comment](https://github.com/pmatos/s11/pull/254#issuecomment-4478527975)):

**Suggestion 1 (`X86Mutator::new` doc comment).** Fixed in `1f5eea0`.
- Changed: rephrased the doc to make explicit that `mode` is consumed at construction (for the Mode32 R8-R15 filter) and not retained as a field.
- Validation: `cargo fmt -- --check`, `cargo build`, `cargo test --bin s11` (820 passing).

**Suggestion 2 (builder consistency in `check_equivalence_x86_for_search`).** **Deferred** — tracked as #260.
- Reason: the AArch64 sibling `EquivalenceConfig` exposes its setter as `timeout()`, not `smt_timeout()`. Picking a name and adding a builder method here without first deciding the cross-ISA naming convention would create new drift. The right shape is a follow-up PR that resolves the naming, adds the builder, and converts every caller (x86 + AArch64) consistently. The current direct-field writes are correct as-is, just stylistically inconsistent.

**Suggestion 3 (stricter regression test).** **Deferred (diminishing returns)** — tracked as #261.
- Reason: you noted this is "not a blocker." The functional regression for item 8 is covered. A true pool-invariant test needs either a negative-case fixture (target with no 1-instruction equivalent under the narrowed pool) or exposing internals — design work that doesn't fit this cleanup PR's scope.

**Coverage CI failure (separate from your review, mentioning here for completeness).** **Deferred** — tracked as #259.
- The `coverage` job fails on two AArch64 enumerative-search timing tests (`respects_available_registers`, `search_honors_timeout`) that pass locally and in the regular `test` job. They're sensitive to `cargo llvm-cov`'s 3-5x instrumentation slowdown. The slack comment at `search.rs:574-578` already acknowledges this is CI-load-dependent. Unrelated to this PR's x86 cleanup scope.

_This triage was made by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.x)._

pmatos:
**Already handled (no change)** — thanks for confirming the triage. All four items are now closed (1 fix in `1f5eea0`, 3 deferrals tracked as #259, #260, #261). The remaining `coverage` job failure is the persistent llvm-cov timing flake tracked in #259 — unrelated to this PR's scope.

Issue #77, opened 2026-05-12
ISA: generalise AArch64/x86/RISC-V backends onto the trait surface


pmatos:
Follow-up to #71.

The traits in `src/isa/traits.rs` (`ISA`, `InstructionType`, `OperandType`, `RegisterType`, `ConcreteExecutor`, `SymbolicExecutor`, `CostModel`, `InstructionGenerator`, `Assembler`) were designed to abstract over multiple backends, but in practice **no backend currently bypasses the trait**:

- AArch64 works directly against `crate::ir::Instruction` everywhere in `src/semantics/`, `src/search/`, `src/assembler/`, `src/elf_patcher/`, `src/main.rs`.
- RISC-V exists only at the trait-impl layer (`src/isa/riscv.rs`); the CLI rejects it (`main.rs:1212-1217`).
- x86 (from #71) ships its own parallel pipeline next to AArch64: `concrete_x86.rs`, `smt_x86.rs`, `cost_x86.rs`, `candidate_x86.rs`, `assembler/x86.rs`, `optimize_elf_binary_x86` in main.rs.

The parallel-pipeline choice was deliberate for #71 to keep the diff reviewable and avoid regressing AArch64. The next step is to **collapse** the duplication.

## What needs doing

1. **Make `src/semantics/` generic over `InstructionType`**. Currently `concrete.rs` matches on `crate::ir::Instruction` exhaustively. The trait surface needed for the semantics is closer to "give me a sum type of 'arithmetic op + condition op + ...' that the interpreter dispatches over" — designing that interface is the bulk of the work.

2. **Decide the width-parameter story**. AArch64 is fixed 64-bit, RISC-V is 32 or 64, x86 is 32 or 64. The `ISA` trait already exposes `register_width()`; the SMT path needs to thread it from there into `MachineState::new_symbolic(prefix)` (currently AArch64-hardcoded to 64 in `smt.rs:69-86`).

3. **`src/search/` parameterisation**. The `SearchAlgorithm` trait in `src/search/mod.rs` takes `&[Instruction]`. After (1), it should take `&[I]` for some `I: InstructionType`. Every concrete searcher (`StochasticSearch`, `SymbolicSearch`, parallel coordinator) becomes generic.

4. **`ElfPatcher`** already has a `DetectedArch` tag (#71); its alignment/NOP behaviours are dispatched from that. This is the simplest part — no further trait work needed.

[...]

PR #255, opened 2026-05-18
opt: build ElfPatcher once at dispatch (#88)

pmatos:
## Summary

Closes #88. Removes the double `fs::read` + `ElfBytes::minimal_parse` that
happened on every `s11 opt` invocation: the CLI arm called
`detect_cli_arch_from_elf`, then each per-arch helper called
`ElfPatcher::new(path)` internally.

After this PR the Opt arm builds the `ElfPatcher` once and threads `&patcher`
into both `optimize_elf_binary` (AArch64) and `optimize_elf_binary_x86`.

## Changes

- `src/main.rs`
  - New `impl From<DetectedArch> for CliArch` next to the `CliArch` enum so
    the Opt arm can derive the CLI-facing arch from the patcher without a
    second parse.
  - Opt arm replaces `detect_cli_arch_from_elf(&binary)` with
    `ElfPatcher::new(&binary)`; the existing `--arch` cross-check + RISC-V
    rejection are preserved.
  - Both `optimize_elf_binary` and `optimize_elf_binary_x86` take
    `patcher: &ElfPatcher` as the first parameter; the internal
    `ElfPatcher::new(path)?` calls are gone. `path` stays for the log lines
    and output-filename construction.
  - `detect_cli_arch_from_elf` is unchanged — the Disasm arm still uses it.
- `src/elf_patcher/mod.rs`
  - New test `elf_patcher_does_not_reread_file_after_construction` builds
    a minimal x86-64 ELF, constructs the patcher, removes the source file,
    and asserts `arch()` / `validate_address_window` /
    `get_instructions_in_window` still succeed. Pins the in-memory contract
    the refactor relies on; any regression that re-introduces a disk read
    breaks this test loudly.

The fix applies to both architectures even though the issue title called
out x86 — the same double-parse existed on the AArch64 path and a half-fix
would have forced an asymmetric dispatch.

## Test plan

- [x] `./ci_check.sh` clean (fmt + build + 821 unit tests + 20 integration
      tests + `test_all.sh`)
- [x] New `elf_patcher_does_not_reread_file_after_construction` test passes
- [x] All existing integration tests in `tests/integration/opt_test.rs`
      pass unchanged (subprocess-based, so they don't care how many times
      the ELF is parsed internally — they only assert observable CLI
      behavior, which this PR preserves)

### DIFF
--- a/src/elf_patcher/mod.rs
+++ b/src/elf_patcher/mod.rs
@@ -563,4 +563,47 @@ mod tests {
             "second pad should be the canonical 8-byte Intel NOP",
         );
     }
+
+    #[test]
+    fn elf_patcher_does_not_reread_file_after_construction() {
+        // Pins the invariant the issue-88 dispatch refactor relies on:
+        // once an ElfPatcher is constructed, every accessor it exposes serves
+        // data from the in-memory buffer rather than reopening the file.
+        // Callers (the `s11 opt` dispatch) can therefore construct the patcher
+        // once and thread it into the per-arch helpers without paying for a
+        // second `fs::read` + `ElfBytes::minimal_parse`.
+        use crate::test_utils::TempFile;
+
+        let text_vaddr: u64 = 0x100000;
+        let text_bytes = [0xc3u8; 8];
+        let elf_bytes = build_minimal_x86_64_elf(&text_bytes, text_vaddr);
+
+        let input = TempFile::new_bytes("s11-elf-no-reread", "elf", &elf_bytes);
+        let saved_path = input.path().to_path_buf();
+        let patcher = ElfPatcher::new(&saved_path).expect("patcher should accept minimal ELF");
+
+        std::fs::remove_file(&saved_path).expect("remove input before exercising patcher");
+        assert!(
+            !saved_path.exists(),
+            "precondition: input file removed so any disk read would fail",
+        );
+
+        assert_eq!(patcher.arch(), DetectedArch::X86_64);
+
+        let window = AddressWindow {
+            start: text_vaddr,
+            end: text_vaddr + 8,
+        };
+        let section = patcher
+            .validate_address_window(&window)
+            .expect("validate should not reopen the file");
+        assert_eq!(section.virtual_addr, text_vaddr);
+
+        let bytes = patcher
+            .get_instructions_in_window(&window)
+            .expect("get_instructions should not reopen the file");
+        assert_eq!(bytes, text_bytes.to_vec());
+
+        // TempFile::drop tolerates a missing file (test_utils.rs:33-37).
+    }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,7 +17,7 @@ mod test_utils;
 mod validation;
 
 use assembler::AArch64Assembler;
-use elf_patcher::{AddressWindow, ElfPatcher, parse_hex_address};
+use elf_patcher::{AddressWindow, DetectedArch, ElfPatcher, parse_hex_address};
 use ir::instructions::split_terminator;
 use ir::{Instruction, Register};
 use search::config::{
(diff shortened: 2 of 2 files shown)


pmatos:
@claude Regarding your review comment (https://github.com/pmatos/s11/pull/255#issuecomment-4478524180):

**Issue 1 — stale `detect_cli_arch_from_elf` reference above `optimize_elf_binary`.** Fixed in `8a87189`.

Changed: replaced the multi-paragraph `Issue #77 stage 1 step 15` block with a short note explaining the helper's new contract (receives a pre-built `ElfPatcher` from the Opt arm) and an updated forward-looking pointer to the future stage-2 step-20 merge — which now naturally branches on `patcher.arch()` instead of `detect_cli_arch_from_elf`.
Validation: `cargo fmt -- --check`, `cargo build`.

**Issue 2 — `elf_patcher` vs `patcher` parameter name mismatch.** Fixed in `e00194a`.

Changed: renamed the parameter in `optimize_elf_binary` from `elf_patcher` to `patcher` (matching `optimize_elf_binary_x86`) and updated the four internal references inside the function body. The `mod elf_patcher;` declaration and `use elf_patcher::…` import at the top of `src/main.rs` are unchanged — those are module-path references, not the parameter.
Validation: `cargo fmt -- --check`, `cargo build`, full `./ci_check.sh` after both commits (fmt + build + 821 unit tests + 20 integration tests + `test_all.sh`).

**Issue 3 — pinned `:33-37` line range in test comment.** **Deferred for separate discussion** — the two AI reviewers disagreed on whether this belongs in this PR (Claude Opus 4.7 classified it as a `style-preference` REJECT since the project's `CLAUDE.md` does not ban pinned line refs and other comments in the codebase do pin them occasionally; Codex GPT-5.x classified it as a small in-scope FIX). Defaulted to DEFER per the autofix dual-evaluator rule so a human can decide whether the project wants a broader no-line-pin convention rather than landing a one-off edit.

Tracked as #262 (https://github.com/pmatos/s11/issues/262).

_This triage was performed by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.x). If you disagree with the deferral, please reply and we'll re-evaluate._

pmatos:
Regarding the codecov report (https://github.com/pmatos/s11/pull/255#issuecomment-4478531296):

**Not an issue** — automated coverage status report, not actionable PR feedback. Project coverage moved 94.21% → 94.23% (above the project gate) and patch coverage is 97.22%. The single missing line is the `unwrap_or_else` error branch on the new `ElfPatcher::new(&binary)` call in the Opt arm, unreachable from tests without a malformed-ELF fixture. No action.

_This assessment was made by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.x). If you disagree, please reply and we'll re-evaluate._

pmatos:
@claude Regarding your second-round review (https://github.com/pmatos/s11/pull/255#issuecomment-4478636396):

**RISC-V dead-arm bug claim.** **Deferred (scope-creep).** The mechanic claim is correct — the `Riscv32 | Riscv64` arm of `match cli_arch` at `src/main.rs:1586` is unreachable. The framing is slightly off, though: that arm was already unreachable **before** this PR. The old `detect_cli_arch_from_elf` returned `Err("Unsupported architecture (e_machine: ...)")` for any non-AArch64/x86 `e_machine`, so `detected_arch` and `cli_arch` were already constrained to `Aarch64 | X86_64 | X86_32`. The "friendly" RISC-V message has never been printed in practice, and the error string template from `ElfPatcher::new` is the same `"Unsupported architecture (e_machine: {})"` as the old function — no UX regression.

Cleanup (your option (a) or (b)) is worthwhile but unrelated to "build `ElfPatcher` once at dispatch". Tracked as #265 (https://github.com/pmatos/s11/issues/265).

**Disasm arm double-read.** **Deferred (scope-creep)** — you already self-deferred this ("Out of scope for this PR, but worth an issue for the follow-up"). Tracked as #266 (https://github.com/pmatos/s11/issues/266).

**Other observations** (positive notes on test coverage, the `From<DetectedArch>` totality comment, and the roadmap-comment cleanup): no action needed — those are confirmations of what's already in the PR.

_This triage was performed by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.x); both agreed on the DEFER outcomes above. If you disagree, please reply and we'll re-evaluate._

pmatos:
Regarding your re-verification comment (https://github.com/pmatos/s11/pull/255#issuecomment-4478624850):

Thanks for re-verifying the fixes. **No action needed** — the second-round review (https://github.com/pmatos/s11/pull/255#issuecomment-4478636396) raised follow-ups that are being handled separately (see the reply above linking #265 and #266).

_This acknowledgement was made by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.x)._

PR #270, opened 2026-05-18
Stabilise enumerative search timing tests under CI load

pmatos:
## Summary
- Override the SMT solver timeout in `search_honors_timeout` to 200ms (was the 30s default) so a single in-flight Z3 query can't dominate the worker-drain wall-clock when the 50ms app-timeout fires. Raise the wall-clock ceiling from 5s to 10s for `cargo-llvm-cov` headroom.
- Raise the internal search timeout in `respects_available_registers` from 10s to 60s so `cargo-llvm-cov`'s 3-5x instrumentation slowdown doesn't expire the budget before the search finds the rewrite.

## Why
The two tests have been flaking across both the `Test` and `Coverage` workflows for weeks. Five open issues track this same flake class. The commit on `main` closes #209 and #183 directly via the message; this PR description closes the remaining three umbrella tickets, all of which describe the same two tests:

Closes #179
Closes #199
Closes #259

## Root cause
Both tests assert wall-clock or completion bounds on a rayon-parallel enumerative search. Workers continue past the stop signal until their current candidate finishes. For `search_honors_timeout`, a candidate that passes the random pre-filter goes to SMT with `solver_timeout` defaulting to 30s — that single query can stall the drain for seconds, breaching the 5s ceiling. For `respects_available_registers`, llvm-cov instrumentation can stretch the 10s search budget so the test never reaches the "found an optimization" assertion.

## Test plan
- [x] `cargo test --bin s11 search::enumerative::search::tests::search_honors_timeout` passes locally
- [x] `cargo test --bin s11 search::enumerative::search::tests::respects_available_registers` passes locally
- [x] `./ci_check.sh` green
- [x] 10 consecutive repeat-runs of both tests pass in <0.5s each
- [ ] CI `Test` workflow passes
- [ ] CI `Coverage` workflow passes (the real proof under llvm-cov instrumentation)

## Out of scope (deliberately)
- Restructuring the search loop to check stop mid-SMT (would need Z3-side interruption).
- Env-var-based CI timeout multiplier — no existing pattern in the codebase and bumping numbers fixes both tests.
- `#[ignore]`-ing either test — loses test signal.

### DIFF
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -402,12 +402,7 @@ mod tests {
         ];
         let live_out = LiveOut::from_registers(vec![Register::X0]);
 
-        // 60s (was 10s) is generous enough to absorb cargo-llvm-cov's 3-5x
-        // instrumentation overhead — under coverage the 10s budget could
-        // expire before the search found the rewrite, failing the
-        // `expect("expected an optimization")` below. The test is asserting
-        // a correctness property (output respects register restrictions),
-        // not search speed, so a loose upper bound is the right knob.
+        // 60s budget absorbs cargo-llvm-cov's 3-5x instrumentation overhead.
         let config = SearchConfig::default()
             .with_registers(vec![Register::X0, Register::X1])
             .with_immediates(vec![0, 1])
@@ -568,14 +563,12 @@ mod tests {
         ];
         let live_out = LiveOut::from_registers(vec![Register::X0, Register::X2]);
 
-        // Override the SMT solver timeout (default 30s) to 200ms so a single
-        // in-flight Z3 query when the 50ms app-timeout fires can't dominate
-        // the worker-drain wall-clock. This is what made the wall-clock
-        // assertion below flake on busy CI runners and under cargo-llvm-cov.
+        // cores=1 + 200ms SMT cap bound post-timeout drain deterministically.
         let config = SearchConfig::default()
             .with_registers(vec![Register::X0, Register::X1, Register::X2])
             .with_immediates(vec![0, 1])
             .with_timeout(std::time::Duration::from_millis(50))
+            .with_cores(Some(1))
             .with_symbolic(
                 SymbolicConfig::default().with_timeout(std::time::Duration::from_millis(200)),
             );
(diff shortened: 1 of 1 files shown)


pmatos:
@claude[bot] Regarding your review — triaged via dual-evaluator (Claude Opus 4.7 + Codex GPT-5.5), four outcomes, one combined commit `6b48c26`:

**1. Comment style violates CLAUDE.md** — **Fixed in `6b48c26`.**
Both block comments on the modified hunks collapsed to one-liners per CLAUDE.md's "one short line max".

**2. 200x slack ratio is very wide** — **Fixed in `6b48c26`.**
Your prediction landed: a coverage run on the parent commit re-failed with `search ran for 11.708777868s, expected < 10s under a 50ms timeout`. Switched `search_honors_timeout` to `.with_cores(Some(1))` so post-timeout drain is bounded by one in-flight candidate × 200ms SMT cap ≈ 1s typical under llvm-cov; the 10s ceiling is now generous headroom rather than a tight target. The rayon parallelism + timeout path is still covered by sibling tests `cores_parallel_finds_same_optimization` and `cores_one_uses_private_pool_not_global`.

**3. Pre-existing `unwrap_or(5s)` inconsistency** — **Deferred (out of scope).**
Real oddity but predates this PR and reconciling it correctly (pick 5s/30s, audit every `SymbolicConfig` constructor, add a regression test) is its own change. Tracked as #281 (https://github.com/pmatos/s11/issues/281).

**4. Slow failure path in `respects_available_registers`** — **No action taken (acceptable trade-off).**
You explicitly called it "acceptable trade-off — just flagging", which matches our reasoning: the test asserts a correctness property, not search speed, so a loose 60s upper bound is the right knob.

_Outcomes determined by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.5). If you disagree on any item, please reply and we'll re-evaluate._

pmatos:
@claude[bot] Regarding your follow-up review — triaged via dual evaluation (Claude Opus 4.7 + Codex GPT-5.5), three REJECTs, no code change. CI is green and the PR has reached fixed point.

**Obs 1: `.with_timeout(60s)` is redundant with `SearchConfig::default()`** — **Style preference (no change).**
Factually correct — `SearchConfig::default()` does set `timeout: Some(Duration::from_secs(60))` at `src/search/config.rs:354`. We're keeping the explicit call: it documents test intent decoupled from the mutable default. If a future change tightens `SearchConfig::default()` (e.g. to 30s for production CLI runs), this test's bound shouldn't silently drift along with it.

**Obs 2: Comment is "slightly cryptic"; expand to two lines** — **Style preference (no change).**
The current one-liner at `src/search/enumerative/search.rs:566` satisfies CLAUDE.md's "one short line max". Your proposed two-line replacement would reintroduce the multi-line comment style your *previous* review correctly flagged as a violation — this commit deliberately shrank both blocks. The terser line covers the same mechanism (`cores=1 + 200ms SMT cap`) without violating the rule.

**Obs 3: `with_cores(Some(0))` edge case is handled** — **No action taken (informational).**
You explicitly flagged this as "fine and unambiguous"; no ask attached. Coercion path is documented at `src/search/config.rs:394-399` and implemented via `n.max(1)` in `src/search/enumerative/search.rs:244-250`.

Thanks for the LGTM. Merging once you do not have further blocking concerns.

_Outcomes determined by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.5). If you disagree on any item, please reply and we'll re-evaluate._

PR #272, opened 2026-05-18
Drop redundant type-name prefix from ParseLiveOutRegistersError Display (closes #80)

pmatos:
## Summary

- `ParseLiveOutRegistersError`'s `Display` impl previously wrote `"ParseLiveOutRegistersError: <msg>"`, which the CLI callers in `src/main.rs:1348,1413` already wrapped as `"invalid live-out: ..."` — producing a doubled prefix like `invalid live-out: ParseLiveOutRegistersError: invalid register name: 'foo'`.
- ADR-0006 §2 commits to `"invalid live-out: ..."` as the sole user-visible CLI prefix, and no in-tree consumers (tests, README, TUTORIAL, scripts) depend on the old `Display` text. The type name remains available via `Debug`.
- `Display` now writes just the message body. A regression test pins the new format and a one-line resolution note has been added to ADR-0006 so the choice is discoverable.

Closes #80.

## Test plan

- [x] `cargo test --bin s11 validation::live_out` — 51/51 pass, including new `display_renders_message_without_type_prefix`
- [x] `./ci_check.sh` — fmt + build + unit + integration green

### DIFF
--- a/docs/adr/0006-live-out-cli-grammar.md
+++ b/docs/adr/0006-live-out-cli-grammar.md
@@ -42,3 +42,5 @@ ADR-0004 §5 commits to eventually replacing `LiveOut` and `X86LiveOutMask` with
 - The bit lives on `EquivalenceConfig` and (in scaffolding) on `LiveOutMask<R>`. When the mask migration of ADR-0004 §5 lands, this ADR's CLI grammar should remain stable but the parser will return `(LiveOutMask<Register>, ())` — a single-return shape — and the explicit `bool` plumbing in `run_equiv` will collapse.
 
 **Reversibility:** high for the per-flag rev — the `n`/`z`/`c`/`v` tokens are rejected with a "reserved" message today, so adding their semantics later is a strict superset of the current grammar. Reversibility low for the parser function itself: it becomes load-bearing for both CLI subcommands.
+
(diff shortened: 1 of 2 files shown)


pmatos:
@claude[bot] thanks for the review. Triaged your three minor suggestions (the fix itself is unchanged):

**Ask A (CLI-level integration test) — Deferred (diminishing returns).** The new unit test pins the type's contribution and the `"invalid live-out: "` CLI prefix is a one-line `format!` in two well-known sites (`src/main.rs:1348,1413`) covered by ADR-0006 §2. A round-trip CLI test is reasonable future hardening but not required here. Tracked as #274 (https://github.com/pmatos/s11/issues/274).

**Ask B (`message` field visibility) — Deferred (out of scope).** You correctly flagged this as a pre-existing follow-up. The doubled-prefix fix shouldn't carry an unrelated visibility change. Tracked as #275 (https://github.com/pmatos/s11/issues/275).

**Ask C (`test_` naming convention) — Style preference (no change).** You explicitly framed this as "either standardise on the prefix or leave it" and noted the new name follows the more modern Rust convention. There's no project rule mandating the `test_` prefix, renaming the new test alone would worsen consistency, and renaming surrounding tests is out of scope for this PR.

_This assessment was made by two independent AI reviewers (Claude Opus 4.7 and Codex GPT-5.5). If you disagree, please reply and we'll re-evaluate._

PR #326, opened 2026-05-21
feat(x86): complete cmov candidate and capability coverage

### LINKED ISSUE #74: x86: add CMOVcc / Jcc and symbolic EFLAGS modelling
Follow-up to #71.

The minimal core ISA in #71 includes CMP but no instructions that read EFLAGS. As a result the symbolic interpreter (`src/semantics/smt_x86.rs`) models CMP as a no-op — the same compromise the AArch64 path makes at `src/semantics/smt.rs:201-204`. The empirical EFLAGS comparison in the fast path (`run_fast_path_x86` in `equivalence.rs`) catches CMP-shortening soundness regressions but the symbolic path remains incomplete.

To meaningfully extend x86 we need:

1. **CMOVcc** (`cmove`, `cmovne`, `cmovl`, `cmovg`, etc.): reads EFLAGS, conditionally writes destination. Needs `X86Instruction::Cmov { rd, rs, cond }` plus an `X86Condition` enum.

2. **Jcc** (`je`, `jne`, `jl`, `jg`, ...): reads EFLAGS, branches. Branches are out of scope for the within-window optimiser, but reasoning about a window's effect on EFLAGS becomes relevant if surrounding code reads them — needed for sound `--live-out flags`.

3. **Symbolic EFLAGS plumbing in `smt_x86.rs`**:
   - Extend `MachineStateX86` with per-flag Z3 `Bool` constants (cf/pf/af/zf/sf/of).
   - `apply_instruction` for arithmetic/logic ops computes and binds those Bool expressions.
   - CMP becomes a non-no-op: it updates the symbolic flag bools.
   - Equivalence comparator extends to assert flag equality when the live-out mask declares flags live.

## Side benefits

- Once symbolic EFLAGS is modelled, the fast-path EFLAGS heuristic in `check_equivalence_x86` becomes a redundant defence-in-depth check rather than the only safety net.
- AArch64 has the same gap (NZCV unmodelled symbolically); the design here can inform a parallel fix on that side.

pmatos:
Summary:
- include CMOVcc in x86 trait/random candidate generation while keeping Jcc out of search pools
- document x86 CMOVcc as rewritable and Jcc as a fixed terminator in the capability matrix
- pin docs/parser/generator coverage for the expanded x86 issue #74 surface

Validation:
- cargo fmt --all
- cargo test --all
- ./ci_check.sh

Closes #74

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,261 +1,103 @@
-# Plan — issue #242: Parallel statistics mislabel symbolic workers and drop most worker metrics
+# Plan — issue #74: x86 CMOVcc, Jcc, and symbolic EFLAGS
 
-## 1. Problem restated
+## 1. Problem Restated
 
-`WorkerMessage::Finished` (src/search/parallel/channel.rs:29-33) only ships `candidates_evaluated`, and the coordinator's `Finished` arm (src/search/parallel/coordinator.rs:150-158) unconditionally synthesises a fresh `SearchStatistics::new(Algorithm::Stochastic)` for every worker, copying only that single field. The symbolic worker — which is always `worker_id == 0` in hybrid mode (src/search/parallel/coordinator.rs:219, src/search/parallel/coordinator.rs:251-272) — therefore appears as `Algorithm::Stochastic` in `worker_statistics`, and `total_statistics` is the sum of one field across workers. Every other stat (`smt_queries`, `smt_equivalent`, `candidates_passed_fast`, `smt_elapsed`, `iterations`, `accepted_proposals`, `improvements_found`, `original_cost`, `best_cost_found`) is silently zero. Because the hybrid CLI path prints `result.total_statistics` (src/main.rs:746), user-visible hybrid reports show 0 SMT/improvement/cost metrics even when a worker has provably found and verified an optimization.
+The x86 backend can already optimize straight-line integer instructions, but CMP is only soundly useful when the backend also models the EFLAGS it writes and the instructions that read those flags. This issue closes that gap by treating CMOVcc as a rewritable flag-reading instruction, treating Jcc as a fixed trailing terminator whose branch outcome observes the prefix's final EFLAGS, and making concrete plus SMT equivalence reject register or EFLAGS divergence whenever flags are live or a pinned flag-reading terminator makes them observable. Current checkout already contains much of this shape; the implementation stage should complete the remaining gaps and normalize the stale surfaces rather than starting from a blank branch.
 
-Fix scope: make `Finished` carry the full per-worker `SearchStatistics` (including its `algorithm` field), and have `run_coordinator` aggregate fields field-by-field rather than synthesising stats from scratch. Also: when an improvement wins, `best_result.statistics` should reflect the actual winning worker's stats, not a hollow `SearchStatistics::new(algorithm)`.
+## 2. Files to Touch
 
-## 2. Files to touch
+No `crates/`, `compiler/`, or `docs/spec/` directories exist in this `s11` checkout, so there is no Rust/self-hosted compiler split and no `docs/spec/*.md` update for this issue. The authoritative docs are `CLAUDE.md`, `docs/capability.md`, and `docs/adr/*.md`.
 
-Production:
-- `src/search/parallel/channel.rs` — extend `WorkerMessage::Finished` payload (add `algorithm: Algorithm`, `statistics: SearchStatistics`); drop the now-redundant standalone `candidates_evaluated` field (the value is `statistics.candidates_evaluated`).
-- `src/search/parallel/coordinator.rs` — three call sites:
-  - `run_symbolic_worker` (lines ~244-276): send `result.statistics` (already carries `Algorithm::Symbolic`) instead of just `candidates_evaluated`.
-  - `run_stochastic_worker` (lines ~279-315): same — send `result.statistics`.
-  - `run_coordinator` (lines ~91-208): change the `Finished` arm to push the received `(worker_id, stats.algorithm, stats)` directly; track the winning worker_id when an `Improvement` wins so `best_result.statistics` can be filled with that worker's final stats; rewrite the aggregate `total_stats` to sum/min relevant fields instead of synthesising. Update the `Improvement` arm so `best_result` is populated *only* with sequence + worker_id + cost — defer assigning `statistics` until all workers have reported.
+Production and docs:
 
-Tests:
-- `src/search/parallel/coordinator.rs` `#[cfg(test)] mod tests` — add the new hybrid regression test the issue asks for (see slice 1) and update the existing tests that assert the dropped-Finished invariant.
-- `src/search/parallel/channel.rs` `#[cfg(test)] mod tests` — `test_create_channels` constructs a `WorkerMessage::Finished` literally and pattern-matches on its fields; update both the construction and the destructure to the new shape.
+- `src/isa/x86.rs` — `X86Condition`, `X86Instruction::{Cmov,Jcc}`, destination/source/side-effect/terminator behavior, `FlagsAnalysis::{modifies_flags,reads_flags}`, stochastic `X86Mutator`, and `X86InstructionGenerator` opcode coverage. Clean up stale "14 variants" / "7 mnemonics" comments and tests.
+- `src/parser/x86.rs` — CMOVcc and Jcc suffix parsing, GAS alias normalization, numeric target validation for Jcc, x86 assembly parser coverage.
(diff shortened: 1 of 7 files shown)


pmatos:
@claude Regarding your code review — triaged each of the three suggestions with two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Outcomes below.

### 1. Magic number `15u32` in `X86Mutator::random_instruction` — ✅ Fixed in `1955ffa`

Both reviewers agreed (FIX). Changed: `src/isa/x86.rs:768` now reads `rng.random_range(0..u32::from(X86_REWRITABLE_OPCODE_COUNT))` instead of the hard-coded `0..15u32`, so `X86Mutator::random_instruction` and `X86InstructionGenerator::generate_random` share a single source of truth for the rewritable opcode-pool size and can't drift if the pool grows again.
Validation: `cargo fmt`, `cargo build`, `cargo test --lib` (158 x86 unit tests incl. `x86_generator_random_within_opcode_range` and the mutator suite), and the `x86_support_is_visible_in_public_docs` integration test — all green.

### 2. `X86_SUPPORTED_MNEMONICS` conflates rewritable instructions and fixed terminators — **Deferred for separate discussion** — tracked as #341

The two reviewers split on whether this belongs in *this* PR (one FIX, one DEFER), and you offered two distinct designs (split off `X86_FIXED_TERMINATORS` vs. rename to `X86_KNOWN_MNEMONICS` + comment). The conflation is real and the AArch64 two-constant split is the cleaner pattern, but the sole current consumer (`tests/integration/docs_capability.rs:168`) only iterates the list for docs-presence assertions and never feeds candidate generation — so the "Jcc treated as a candidate" risk is hypothetical today. Filed as a focused follow-up that can pick one design rather than churn it here. Tracked as #341 (https://github.com/pmatos/s11/issues/341).

### 3. `has_side_effects` semantics for CMOV — documentation gap — **Already handled (no change)**

The clarification you asked for is already present in the implementation. `src/isa/x86.rs:389-392` reads: "MOV / CMOV / Jcc do not write EFLAGS (CMOV and Jcc read them, but reading is not a side effect on observable state). Every other variant sets or clobbers flag bits, which is observable state beyond the destination register." That states exactly that `side_effects` means "writes EFLAGS" and that CMOV reads-but-doesn't-write — so no further change is needed.

_The assessments above were made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree with the deferral or the no-change items, please reply and we'll re-evaluate._

pmatos:
@claude Thanks for the second pass. Triaged all five items with two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5); both agreed on every verdict.

### 1. `X86_SUPPORTED_MNEMONICS` mixes rewritable instrs with a fixed terminator — **Deferred (already tracked)** — #341

Same point as the first review's suggestion 2; already filed as #341 (https://github.com/pmatos/s11/issues/341) for a focused split (`X86_FIXED_TERMINATORS`) vs. rename decision. The sole consumer today (`tests/integration/docs_capability.rs:168`) is docs-presence only, so there's no candidate-generation hazard in this PR.

### 2. Clarify that `X86_REWRITABLE_OPCODE_COUNT = 15` counts families, not instances — ✅ Fixed in `18149fd`

Both reviewers agreed (FIX). Changed: added a comment above `src/isa/x86.rs:974` noting the constant counts rewritable opcode *families* (7 reg-reg + 7 reg-imm + CMOVcc = 15) and that CMOVcc is one family even though `generate_all` expands it across all 16 `X86Condition::ALL` variants per register pair.
Validation: `cargo fmt` + `cargo build` (comment-only change, no behavior impact).

### 3. Soundness — confirm CMOV concrete/SMT/equivalence semantics exist — **Already handled (no change)**

Confirmed against the code; this work landed in earlier commits on this branch (the "symbolic-eflags-modelling" half of the branch), so adding CMOV to the candidate pools here is sound:
- **Concrete:** `src/semantics/concrete_x86.rs:45-51` does `if state.get_flags().evaluate(*cond) { set_register(rd, rs) }` — it evaluates the condition against the *incoming* EFLAGS, not a zero default.
- **SMT:** `src/semantics/smt_x86.rs:315-320` lowers it symbolically — `x86_condition_to_smt(*cond, state.get_flags())` then `pred.ite(rs_val, rd_old)`; the flag BVs are left untouched (test at `smt_x86.rs:745`).
- **Equivalence:** flag-reading sequences are detected via `reads_flags_before_writing` and the x86 checker randomizes the five tracked input EFLAGS bits (`equivalence.rs` ≈900-911); the SMT path uses a fully symbolic initial state, and `flags_not_equal_x86` refutes flag divergence when flags are live.

So the "zero flags → CMOVE never fires → spuriously equivalent" failure mode is already guarded. No change needed.

### 4. `cond` sampled unconditionally in `X86Mutator::random_instruction` — **Style preference (no change)**

The unconditional draw is intentional and, importantly, *consistent* with the parallel `generate_random` (`src/isa/x86.rs:1025`), which samples `cond` the same way. The current code is fully deterministic and reproducible as written. Switching to lazy evaluation would change the seeded-RNG sequence (risking existing seeded tests) and desync the two random surfaces unless both change — a stability cost for no correctness gain. Leaving as-is.

### 5. Add a `Jcc` `Display` pinning test — **Out of scope for this PR** — #347

You framed this as a follow-up, and both reviewers agreed (DEFER). `Display` already emits `j{cond} <target>` (`src/isa/x86.rs:424`), a non-parseable placeholder — exactly the safe behavior. Filed a focused test as #347 (https://github.com/pmatos/s11/issues/347).

_The assessments above were made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree with any deferral or no-change item, reply and we'll re-evaluate._

pmatos:
Re: the third `claude-review` pass (comment above) — triaged all five items with two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). No code change was warranted; one new tracking issue filed. (Addressing without an `@`-mention to avoid re-triggering the review bot — this is the convergence reply.)

### 1. Eager `cond` perturbs the RNG stream for non-CMOV opcodes — **Style preference (no change)**

Same point as the second review's item 4, already triaged. The unconditional `cond` draw is consistent across all three random surfaces (`random_instruction`, `generate_random`, `generate_random_x86_instruction`) and the code is deterministic as written. Note the stream necessarily changes vs. pre-PR anyway because the opcode range itself grew (`0..14` → `0..15`) and `Cmov` is newly reachable — so lazy-drawing `cond` would not restore the old stream, it would just add a third inconsistency. No correctness impact; leaving as-is.

### 2. Three parallel opcode dispatch tables — **Deferred** — #348

Confirmed three in-sync copies (`src/isa/x86.rs:766`, `src/isa/x86.rs:1018`, `src/search/candidate_x86.rs:98`). The duplication predates this PR (which only added the CMOVcc arm to the existing third surface); consolidating three dispatchers is a non-trivial refactor outside this PR's scope. Filed as #348 (https://github.com/pmatos/s11/issues/348), which also captures the lighter "add keep-in-sync comments" alternative.

### 3. Parser test coverage is partial (only `cmove`/`je`) — **Already handled (no change)**

The exhaustive coverage already exists, separate from the `x86_ir_recognises_supported_mnemonic_families` smoke test: `src/parser/x86.rs` has `parses_all_canonical_cmov_suffixes` (all 16 CMOV conditions, line 575), `parses_all_canonical_jcc_suffixes` (all 16 Jcc conditions, line 631), `cmov_with_unknown_suffix_errors` (line 611), and `jcc_gas_aliases_normalize_to_canonical_conditions` (GAS aliases like `jnae`→B, `jnz`→NE, line 693). The 16-condition + alias gap the review describes is already closed.

### 4. PLAN.md replacement loses issue #242 plan — **Not an issue (no change)**

Issue #242 is **closed as completed** (2026-05-20), so no live plan was lost. PLAN.md is ephemeral per-issue working state in this repo — git history shows it being swapped per issue (`replace PLAN.md with issue #242 plan` → `refresh issue 74 implementation plan`), and the first review already accepted this as "expected churn given the project workflow." Nothing actionable here.

### 5. `X86_SUPPORTED_MNEMONICS` naming — **Deferred (already tracked)** — #341

Same as item 1 of the second review; tracked in #341 (https://github.com/pmatos/s11/issues/341).

_Assessed by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). All prior FIX items remain committed (`1955ffa`, `18149fd`) and CI is green. This PR is at its fixed point: green CI + every feedback item triaged and answered._

PR #328, opened 2026-05-21
test(candidate): cover random MADD-family generation

### LINKED ISSUE #175: Add test_generate_random_reaches_madd_family
Deferred from #171: add a fixed-seed 5000-draw test in `src/search/candidate.rs` that asserts `generate_random_instruction` still reaches opcode_ids 42..=46 (Madd, Msub, Mneg, Smulh, Umulh) after slot 29 of the random dispatch was promoted from `_ =>` to an explicit arm.

Original feedback by @claude[bot] on PR #171 (https://github.com/pmatos/s11/pull/171#issuecomment-4475677414):

> **4. Minor: MADD-group regression after `_ =>` promotion**
>
> Slot 29 was previously the `_ =>` catch-all; it is now an explicit arm. This is correct. However, there is no new test asserting that `generate_random_instruction` still reaches the MADD group after the refactor. The existing `test_generate_all_instructions_madd_msub_*` tests cover the exhaustive path but not the random path. A single `test_generate_random_reaches_madd_family` on the same pattern would close the gap (even if implicitly it worked before).

**Context:** `src/search/candidate.rs:705-721` (MADD sub-multiplexer at slot 29) and `src/search/candidate.rs:1023-1061` (random-reach test pattern).

**Why deferred:** `diminishing-returns` — slot 29 is reached by other random-walk tests (`test_generate_random_instruction`, `test_generate_random_instruction_emits_bitfield_eventually`) which would also fail loudly if the slot disappeared, and the MADD family is not the subject of this PR. The evaluators disagreed (Claude Opus = DEFER diminishing-returns; Codex = FIX, high confidence) so per the autofix dual-evaluator policy this defaults to DEFER for a follow-up test-coverage pass.

pmatos:
Summary:
- Add a fixed-seed 5,000-draw regression test proving random candidate generation reaches the MADD-family opcode IDs 42..=46.
- Preserve the issue-specific implementation plan in PLAN.md for this reused workspace.
- Make two timing-sensitive smoke tests robust under coverage instrumentation after the first PR run exposed the bench timeout edge.

Closes #175

Verification:
- Temporary red check: forced random slot 29 away from the MADD family and saw the new test fail with `random never produced Madd`.
- `cargo test search::candidate::tests::test_generate_random_reaches_madd_family -- --exact`
- `cargo test bench_support::tests::run_bench_enumerative_records_metrics_for_fusible_target -- --exact`
- `cargo test search::enumerative::search::tests::statistics_aggregate_smt_elapsed -- --exact`
- `./ci_check.sh`
- `cargo llvm-cov --workspace --all-features --lcov --output-path /tmp/s11-issue175-lcov.info`

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,261 +1,68 @@
-# Plan — issue #242: Parallel statistics mislabel symbolic workers and drop most worker metrics
+# Plan - issue #175: Add `test_generate_random_reaches_madd_family`
 
-## 1. Problem restated
+## 1. Problem Restated
 
-`WorkerMessage::Finished` (src/search/parallel/channel.rs:29-33) only ships `candidates_evaluated`, and the coordinator's `Finished` arm (src/search/parallel/coordinator.rs:150-158) unconditionally synthesises a fresh `SearchStatistics::new(Algorithm::Stochastic)` for every worker, copying only that single field. The symbolic worker — which is always `worker_id == 0` in hybrid mode (src/search/parallel/coordinator.rs:219, src/search/parallel/coordinator.rs:251-272) — therefore appears as `Algorithm::Stochastic` in `worker_statistics`, and `total_statistics` is the sum of one field across workers. Every other stat (`smt_queries`, `smt_equivalent`, `candidates_passed_fast`, `smt_elapsed`, `iterations`, `accepted_proposals`, `improvements_found`, `original_cost`, `best_cost_found`) is silently zero. Because the hybrid CLI path prints `result.total_statistics` (src/main.rs:746), user-visible hybrid reports show 0 SMT/improvement/cost metrics even when a worker has provably found and verified an optimization.
+Issue #175 asks for a focused regression test proving that the AArch64 random candidate generator still reaches the multiply-accumulate instruction family after the dispatch slot that used to be the catch-all became an explicit `29 =>` arm. The existing exhaustive generator already covers `Madd`, `Msub`, `Mneg`, `Smulh`, and `Umulh`, but the stochastic path needs its own fixed-seed 5000-draw guard over `generate_random_instruction` so future refactors cannot silently make opcode IDs `42..=46` unreachable.
 
-Fix scope: make `Finished` carry the full per-worker `SearchStatistics` (including its `algorithm` field), and have `run_coordinator` aggregate fields field-by-field rather than synthesising stats from scratch. Also: when an improvement wins, `best_result.statistics` should reflect the actual winning worker's stats, not a hollow `SearchStatistics::new(algorithm)`.
+## 2. Files To Touch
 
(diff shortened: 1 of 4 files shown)


pmatos:
@claude[bot] Regarding your review (the first code review, comment 4506775623):

Triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Outcomes per point:

**PLAN.md inclusion — No change.** `PLAN.md` has been a tracked file since the repo's first commit (`9a17984`) and is deliberately swapped per-issue via dedicated `docs(plan):` commits (`7eb0ba8` → #242, `b2f03b2` → #175); it is intentionally not in `.gitignore`. The #242 plan is not lost — it remains in git history (`git log -- PLAN.md`). Both reviewers agree the swap is by-design here (consistent with your own review #2: "routine workspace hygiene. No concerns."), so no change in this PR.

**Deferred (separate discussion) — merge-strategy claim in the PR body.** You're right that CLAUDE.md (Commit Guidelines, CLAUDE.md:214-216) says nothing about merge vs squash, so "per CLAUDE.md" is unsupported. The two reviewers split on whether to act on it now, so rather than silently drop it I filed a tracking issue: either remove the citation from the body or document an actual merge policy in CLAUDE.md. Tracked as #342 (https://github.com/pmatos/s11/issues/342).

**Style preference (no change) — shared seed `0x66`.** As you noted, this is "a style decision rather than a fix." The shared seed is the established convention across all four `test_generate_random_reaches_*_family` tests; they assert deterministic reachability over one large sample, not statistical independence, so a shared seed is correct by design.

pmatos:
@claude[bot] Regarding your review (the second code review, comment 4506848187):

Triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Thanks for the positive notes on the `statistics_aggregate_smt_elapsed` simplification and on `PLAN.md` — agreed, no action needed there. On the two suggestions:

**Deferred (diminishing returns) — document why 5000 draws is sufficient.** Useful idea, but the new MADD test deliberately mirrors three pre-existing siblings (`mul_div`, `compare`, `csel`) that carry no such comment; adding it to only the new test would make the suite inconsistent. Better done as a uniform sweep over all four — out of scope for this green test-only PR. Tracked as #343 (https://github.com/pmatos/s11/issues/343).

**Deferred (diminishing returns) — gate the bench timeout instead of the unconditional 120s bump.** The concern is legitimate (a 120s ceiling could mask a future regression), but as you said it's "Not a blocker": the test asserts emitted metrics, not a wall-clock budget, and the commented bump is a reasonable choice. A `#[cfg(coverage)]` / env-var gate is a fair refinement worth tracking separately. Tracked as #344 (https://github.com/pmatos/s11/issues/344).

PR #335, opened 2026-06-07
docs: fix mutants summary unittest command

szx19970521:
Fixes #204.

The documented unittest invocation in `scripts/test_mutants_summary.py` currently fails from the repository root because `scripts/` is not placed on `sys.path` for `python -m unittest scripts/test_mutants_summary.py`.

This updates the file header to the working unittest discovery command.

Validation:

```text
python -m unittest scripts/test_mutants_summary.py
# fails before this change with ModuleNotFoundError: No module named 'mutants_summary'

python -m unittest discover -s scripts -p 'test_*.py'
# Ran 28 tests in 0.319s - OK
```


pmatos:
thx

PR #350, opened 2026-06-15
fix(x86): mutate cmp opcode forms

### LINKED ISSUE #190: X86Mutator::mutate_opcode is a no-op for CMP instructions
Deferred from #184: when opcode mutation picks a CMP slot, `X86Mutator::mutate_opcode` returns the instruction unchanged.

Original feedback by @claude bot on PR #184 (https://github.com/pmatos/s11/pull/184#issuecomment-4476263368):

> When a CMP is selected for opcode mutation, the sequence is returned unchanged. The rationale (CMP has no destination so it can't adopt another instruction's shape) is sound, but at minimum `CmpReg ↔ CmpImm` (same semantics, different source operand form) should be a valid opcode mutation. Currently, any sequence whose opcode-mutation slot lands on a CMP wastes that mutation step entirely, which degrades stochastic search quality.

**Context:** `src/isa/x86.rs:719-756` (`X86Mutator::mutate_opcode`).

**Why deferred:** diminishing-returns — search-quality improvement, not a correctness issue. Stochastic search still works; it just wastes ~7% of opcode-mutation steps when the picked slot is a CMP. Worth doing on its own.

pmatos:
Summary:
- Let x86 opcode mutation bridge `CmpReg` and `CmpImm` while preserving the compared register.
- Add opcode-only mutator regressions for both CMP operand forms.

Verification:
- `./build_tests.sh`
- `cargo test --all`
- `./ci_check.sh`

Note: `scripts/full_test.sh` is not present in this s11 workspace; `./ci_check.sh` is the repo-local full gate from `CLAUDE.md`.

Closes #190

### DIFF
--- a/src/isa/x86.rs
+++ b/src/isa/x86.rs
@@ -865,6 +865,12 @@ impl X86Mutator {
     /// its operand shape (reg-reg → reg-reg, reg-imm → reg-imm). CMP
     /// has no rd, so it only swaps between register and immediate CMP
     /// forms.
+    ///
+    /// Note the deliberate asymmetry: the reg-reg and reg-imm groups
+    /// sample from a range that includes the current variant, so they may
+    /// produce an identity mutation. CMP, by contrast, always bridges
+    /// `CmpReg` ↔ `CmpImm`, so a CMP opcode mutation is guaranteed to
+    /// change the form. This is intentional, not an oversight.
     fn mutate_opcode<R: rand::RngExt>(&self, rng: &mut R, sequence: &mut [X86Instruction]) {
         if sequence.is_empty() {
             return;
@@ -1718,7 +1724,9 @@ mod tests {
 
         let mutator = X86Mutator::new(
             vec![X86Register::RBX],
-            vec![5],
+            // Unused by CmpImm → CmpReg (which calls pick_register, not
+            // pick_immediate); a value absent from the target makes that clear.
+            vec![0],
             MutationWeights {
                 operand: 0.0,
                 opcode: 1.0,
@@ -1727,8 +1735,10 @@ mod tests {
             },
             crate::assembler::x86::X86Mode::Mode64,
         );
+        // rn is a non-RAX register so the "rn is preserved" assertion can't be
+        // satisfied coincidentally by the pick_register RAX fallback default.
         let target = vec![X86Instruction::CmpImm {
-            rn: X86Register::RAX,
+            rn: X86Register::RCX,
             imm: 5,
         }];
         let mut rng = ChaCha8Rng::seed_from_u64(7);
@@ -1738,7 +1748,7 @@ mod tests {
         assert_eq!(
             mutated,
             vec![X86Instruction::CmpReg {
-                rn: X86Register::RAX,
+                rn: X86Register::RCX,
                 rs: X86Register::RBX,
             }]
         );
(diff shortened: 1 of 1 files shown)


pmatos:
Implementation run notes:

- `scripts/full_test.sh` is absent in this s11 workspace, so I used the repo-local `./ci_check.sh` from `CLAUDE.md` after `cargo fmt`, `cargo clippy`, `./build_tests.sh`, and `cargo test --all`.
- The issue readiness label present was `ready-for-agent`, not `agent-ready`, so I removed `ready-for-agent` after opening the PR.
- `PLAN.md` was a pre-existing planning artifact in the reused workspace and remains uncommitted; the PR contains only the source/test change.

pmatos:
@claude Regarding your code review (the three "Minor nits") — all three were triaged **FIX** by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5) and are addressed in `7cc965a`.

**Nit 1 — unused immediate pool in the `CmpImm → CmpReg` test.** Fixed. Replaced `vec![5]` with `vec![0]` (plus a comment) so it's obvious the immediate pool is irrelevant to that assertion — the arm calls `pick_register`, not `pick_immediate`.

**Nit 2 — `rn` preservation only exercised for RAX.** Fixed. Switched that test's `rn` to a non-RAX register (`RCX`, input and expected output). This also guards against a subtler trap: `pick_register` falls back to `RAX` on an empty pool, so RAX was a treacherous choice for a preservation assertion. The two tests now pin `rn` for two distinct registers (RAX and RCX), making the "rn is always preserved" invariant explicit.

**Nit 3 — undocumented asymmetry.** Fixed. Added a note to the `mutate_opcode` doc-comment explaining that the reg-reg/reg-imm groups sample from a range that includes the current variant (so they may self-map), whereas CMP always bridges `CmpReg ↔ CmpImm` and never produces an identity mutation — calling out that this is intentional, not an oversight.

Changed: `src/isa/x86.rs` (the two CMP opcode-mutator tests and the `mutate_opcode` doc-comment) — no production-behavior change.
Validation: `cargo fmt -- --check`, `cargo clippy --all -- -D warnings` (clean), `./build_tests.sh`, and `cargo test` — both CMP mutator tests and the full integration suite pass.

pmatos:
@claude Regarding your second-round review (the three non-blocking suggestions, comment #4713140204) — triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Thanks for the re-review and the approve. Outcomes below; no code change this round, to avoid churning an already-approved PR over optional polish.

**Suggestion 1 — reword the `vec![0]` immediate-pool comment: No action taken (style preference).** Both reviewers agreed (REJECT). The existing comment already states the same fact the rewording would — that this arm calls `pick_register`, not `pick_immediate`, so the immediate pool is unused — and you noted the current comment is "good enough." Re-spinning a correct comment would only trigger another review cycle for no semantic gain.

**Suggestion 2 — anchor the `rn`-preservation invariant with a comment: Deferred (diminishing returns).** Test B already documents this near its target, and both tests' assertions encode the invariant directly (`rn` in == `rn` out). The reviewers split on whether a near-assertion comment in Test A adds value, so rather than churn it in here, tracked as #367 (https://github.com/pmatos/s11/issues/367).

**Suggestion 3 — multi-instruction test for random index selection: Deferred (diminishing returns).** Correct observation, but the `rng.random_range(0..sequence.len())` index pick is shared logic used by every mutation operator and already exercised by other tests; a CMP-specific multi-instruction test is low-value here (you flagged it low-risk/non-blocking). Tracked as #368 (https://github.com/pmatos/s11/issues/368).

_This triage was made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree with any outcome, reply and we'll re-evaluate._

PR #354, opened 2026-06-15
refactor(validation): add live-out contract parse error alias

### LINKED ISSUE #180: Rename `ParseLiveOutRegistersError` to broader contract-parse scope
Deferred from #172: `parse_live_out_contract` reuses `ParseLiveOutRegistersError` but the error now also carries grammar/flag-token diagnostics.

Original feedback by @claude[bot] on PR #172 (https://github.com/pmatos/s11/pull/172#issuecomment-4475747275):

> `parse_live_out_contract` returns `ParseLiveOutRegistersError`, which was originally designed for register-only errors. Now it carries messages like `"unknown flag token 'bogus'"`. This isn't wrong — the existing error type is just a `{ message: String }` wrapper — but a reader seeing `ParseLiveOutRegistersError` on a contract-parse call may do a double-take. Options: leave + doc comment, rename to `ParseLiveOutError`, or add a type alias.

**Context:** `src/validation/live_out.rs:11` (struct definition) and call sites in `src/main.rs` `run_equiv` / `run_llm_opt`.

**Why deferred:** diminishing-returns — one evaluator preferred to defer (renaming a public error type has blast radius beyond this PR and is purely aesthetic), while the other preferred to reject outright. With the evaluators disagreeing, the safe path is to file a tracking issue so the decision isn't silently dropped.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.7 and Codex GPT-5.4._

pmatos:
Summary:
- add ParseLiveOutError as the contract-facing alias for --live-out parsing errors
- update parse_live_out_contract tests and signature to use the broader name
- refresh ADR-0006 and the tracked issue plan terminology

Verification:
- cargo fmt --all
- ./build_tests.sh
- cargo test --all (first attempt failed before fixtures existed; rerun passed after ./build_tests.sh)
- ./ci_check.sh (first attempt hit a transient symbolic cancellation test; exact rerun and second full wrapper run passed)
- scripts/full_test.sh was requested by the autonomous contract but is not present in this s11 checkout; ./ci_check.sh runs ./test_all.sh.

Closes #180

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,68 +1,62 @@
-# Plan - issue #175: Add `test_generate_random_reaches_madd_family`
+# Plan: Contract-Scoped Live-Out Parse Error
 
 ## 1. Problem Restated
 
-Issue #175 asks for a focused regression test proving that the AArch64 random candidate generator still reaches the multiply-accumulate instruction family after the dispatch slot that used to be the catch-all became an explicit `29 =>` arm. The existing exhaustive generator already covers `Madd`, `Msub`, `Mneg`, `Smulh`, and `Umulh`, but the stochastic path needs its own fixed-seed 5000-draw guard over `generate_random_instruction` so future refactors cannot silently make opcode IDs `42..=46` unreachable.
+`parse_live_out_contract` now parses the whole `--live-out` contract, including the optional `;nzcv` flag-liveness suffix, so a register-only error name is misleading when diagnostics can say things like `unknown flag token 'bogus'`. In this checkout the original `ParseLiveOutRegistersError` symbol is already gone and the shared wrapper is named `ParseRegisterSetError`, but the contract parser still exposes that register-set-flavored name in its signature and ADR-0006 still has one stale reference to the old name. The implementation should add a contract-facing error name for `parse_live_out_contract` without breaking existing `RegisterSet<Register>::from_str` users.
 
 ## 2. Files To Touch
 
-- `src/search/candidate.rs` - add one unit test in the existing `#[cfg(test)] mod tests`, near the current random reachability tests and reusing `random_opcode_ids(seed, draws)`.
-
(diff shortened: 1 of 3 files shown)


pmatos:
Regarding your automated code review above — triaged all 5 items with two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Outcome per item:

**1. Misleading test name — No action taken.** The name `display_renders_message_without_type_prefix` still accurately describes the assertion: that `Display` (`err.to_string()`) renders only the message body with no `ParseLiveOutError {...}`-style type prefix. Routing through `parse_live_out_contract("x0;bogus")` doesn't change what's verified. More decisively, ADR-0006's resolution note pins this exact path (`validation::live_out::tests::display_renders_message_without_type_prefix`) as the guardrail — renaming would desync the ADR for a stylistic gain.

**2. Test duplication — No action taken.** `display_renders_message_without_type_prefix` and `test_parse_live_out_contract_unknown_flag_rejected` share the input `"x0;bogus"` but assert different contracts: exact full-string `Display` equality vs. a `contains("unknown flag token")` rejection-category check. They're complementary, and the comment requested no concrete change.

**3. Use the alias in the body — Out of scope for this PR (deferred).** Legitimate intent-signaling, but purely stylistic over already-correct, behaviour-preserving code, and the two evaluators split on it. Tracked as #359 so it isn't lost. (Note: `parse_register()` / `RegisterSet::from_str` keep `ParseRegisterSetError` since they raise genuine register-set errors.)

**4. Re-export from `validation/mod.rs` — Out of scope for this PR (deferred).** Reasonable ergonomic addition, but it widens public surface with no current consumer — callers use the full `validation::live_out::` path, and the alias is already public there. Tracked as #360.

**5. ADR date one day ahead — Not applicable.** Today is 2026-06-16, exactly the date in the resolution note (`issue #180, 2026-06-16`), so it isn't future-dated; the 2026-06-15 observation is now stale.

_Items 1, 2, and 5 were assessed by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree, reply and we'll re-evaluate._

PR #369, opened 2026-06-15
fix(search): make x86-32 backend width architectural

pmatos:
## Summary
- Make x86-32 stochastic and symbolic backend width accessors return the ISA marker width instead of SearchConfig::x86_width
- Add regression coverage proving default SearchConfig cannot silently widen X86_32 search to 64 bits
- Update the x86-32 symbolic end-to-end test comment to describe the new invariant

## Verification
- cargo fmt --all
- cargo test --all (after ./build_tests.sh; initial run failed only because fixture binaries were absent)
- ./ci_check.sh
- scripts/full_test.sh is not present in this s11 repository; used ./ci_check.sh per CLAUDE.md

Closes #192

**Merge with  (merge commit, not squash) per CLAUDE.md.**

### DIFF
--- a/src/search/stochastic/backend.rs
+++ b/src/search/stochastic/backend.rs
@@ -110,6 +110,8 @@ pub trait StochasticBackend<I: ISA>: Sized {
 
     /// Width parameter for cost + state masking. Architecture markers own
     /// this width so a mismatched config cannot silently change semantics.
+    /// `config` is retained only for API symmetry; implementations are
+    /// expected to return an architectural constant and must not read it.
     fn width(config: &SearchConfig) -> u32;
 }
 
@@ -449,6 +451,8 @@ mod tests {
             32
         );
 
+        // X86_64 was already correct; this cross-check guards against a
+        // future regression and is not part of the x86-32 bug being fixed.
         assert_eq!(
             <crate::isa::X86_64 as StochasticBackend<crate::isa::X86_64>>::width(
                 &SearchConfig::default().with_x86_width(32),
--- a/src/search/stochastic/backend.rs
+++ b/src/search/stochastic/backend.rs
@@ -108,8 +108,8 @@ pub trait StochasticBackend<I: ISA>: Sized {
         None
     }
 
-    /// Width parameter for cost + state masking. AArch64 returns 64;
-    /// x86 reads `SearchConfig::x86_width` (32 or 64).
+    /// Width parameter for cost + state masking. Architecture markers own
+    /// this width so a mismatched config cannot silently change semantics.
     fn width(config: &SearchConfig) -> u32;
 }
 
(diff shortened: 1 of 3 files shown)


pmatos:
Regarding your code-review summary (the three observations), each triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5):

**1. `SearchConfig::x86_width` / `x86_mode` are now write-only — Out of scope for this PR.** Both reviewers agreed the observation is correct, but it's a pre-existing maintenance hazard whose resolution (remove the fields vs. document them as reserved) touches `config.rs` and call sites beyond this PR's focused three-file scope. Tracked as #372 (https://github.com/pmatos/s11/issues/372).

**2. `width(config:)` trait parameter is unused — Fixed in `ae9819a`.**
Changed: extended the `width()` doc comment on both `StochasticBackend` and `SymbolicBackend` to state the `config` parameter is retained only for API symmetry and that implementations are expected to return an architectural constant and must not read it.
Validation: `cargo fmt --check` clean; `cargo build` compiles (doc comments well-formed).

**3. X86_64 cross-check assertion needs a clarifying comment — Fixed in `ae9819a`.**
Changed: added a comment above the `X86_64::width(&SearchConfig::default().with_x86_width(32)) == 64` assertion in both backend regression tests, marking it as a non-regression cross-check rather than part of the x86-32 bug being fixed.
Validation: both touched unit tests pass (`x86_32_stochastic_width_is_architectural_even_with_default_config`, `x86_32_symbolic_width_is_architectural_even_with_default_config`).

_Triaged and applied automatically by `pm-autofix-pr`. If you disagree with the deferral, reply and we'll re-evaluate._

pmatos:
Regarding your follow-up review (and thanks for the approval) — its two observations, each triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5):

**1. Remove the `config` parameter from the `width()` trait entirely — Out of scope for this PR.** Both reviewers agreed the "retained for API symmetry / must not read it" contract is a smell better resolved by dropping the parameter, but doing so spans both the `StochasticBackend` and `SymbolicBackend` traits, all their impls, and every call site (`mcmc.rs`, `synthesis.rs`, cost / state-masking) — broader than this targeted width bug fix, as you note. Folded into the existing vestigial-config cleanup issue #372 (https://github.com/pmatos/s11/issues/372), which now also tracks the trait-parameter removal.

**2. Reword the X86_64 cross-check comment ("also guards the symmetric X86_64 invariant") — No change.** Both reviewers found the existing comment accurate: the X86_64 assertion verifies that a mismatched `with_x86_width(32)` config cannot alter the architectural 64-bit width, and that was indeed not the original x86-32 defect. The suggested phrasing is equally valid, so this is a stylistic preference with no correctness or maintainability gain — not worth another change in this PR.

_Triaged automatically by `pm-autofix-pr`. If you disagree with either outcome, reply and we'll re-evaluate._

PR #379, opened 2026-06-16
perf(smt): reduce clz formula depth

### LINKED ISSUE #112: Replace 64-deep ITE chain in `bv_clz_64` with binary-search decomposition
Deferred from #109: the current `bv_clz_64` (and indirectly `bv_cls_64` via the sign-fold trick) builds a 64-deep nested ITE chain to compute leading zeros symbolically.

Original feedback by @claude on PR #109 (https://github.com/pmatos/s11/pull/109#issuecomment-4441883408, suggestion 1):

> Each call to `bv_clz_64` inserts 64 nested ITEs into the Z3 formula. CLZ is semantically a "count" operation, and nested ITEs create a large term that Z3's preprocessors don't always simplify well. For sequences with multiple CLZ instructions (or CLS, which calls it twice), equivalence queries could get slow.
>
> A standard alternative is a binary-search decomposition: check the top 32 bits first, then 16, 8, 4, 2, 1 — giving log2(64) = 6 levels of ITE instead of 64.

**Context:** `src/semantics/smt.rs:51-61` (`bv_clz_64`); also reached transitively by `Instruction::Cls` via the `x ^ (x ASR 63)` fold (which calls `bv_clz_64` once per CLS instance).

**Why deferred:** diminishing-returns — no measured slowdown yet. The change is a non-trivial semantic rewrite that deserves its own PR with before/after Z3 timing benchmarks. The current implementation is provably correct (the floor-log2 acceptance test in #109 passes), so this is purely a performance follow-up.

**Suggested approach:** rewrite as a recursive halving — check whether the top half of the bit range is zero; if so, the answer is `(width/2) + clz(bottom_half)`, else `clz(top_half)`. Base case at width 1 collapses to a single ITE. Adds 6 levels of ITE for a 64-bit BV instead of 64.

pmatos:
Summary:
- Replace the 64-step `bv_clz` nested ITE chain with a binary-search decomposition over 32/16/8/4/2/1-bit halves.
- Add a formula-shape regression test plus expanded CLZ/CLS concrete/SMT boundary coverage.
- Add a direct `smt_clz` Criterion benchmark and `just bench-smt-clz` recipe for CLZ/CLS SMT timing.

Benchmark evidence (`cargo bench --bench smt_clz -- --quick`, Criterion middle estimate; old helper measured by temporarily reversing only `src/semantics/smt.rs`):
- old: one_clz 2.5672 ms, four_clz 8.6741 ms, one_cls 3.3232 ms, four_cls 11.422 ms
- new: one_clz 875.73 us, four_clz 3.2310 ms, one_cls 897.27 us, four_cls 3.3663 ms

Verification:
- `cargo test semantics::smt::tests::test_bv_clz_formula_has_logarithmic_ite_depth -- --exact` (red on old helper with nested ite depth 64; green after rewrite)
- `cargo test semantics::smt::tests::test_clz_concrete_smt_parity -- --exact`
- `cargo test semantics::smt::tests::test_cls_concrete_smt_parity -- --exact`
- `cargo test semantics::smt::tests::test_clz_floor_log2_pattern -- --exact`
- `cargo clippy --bench smt_clz -- -D warnings`
- `./build_tests.sh`
- `cargo test --all`
- `./ci_check.sh`

Closes #112

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,3 +52,7 @@ harness = false
 [[bench]]
 name = "algebraic_fusion"
 harness = false
+
+[[bench]]
+name = "smt_clz"
+harness = false
--- a/Justfile
+++ b/Justfile
@@ -96,6 +96,10 @@ bench-phase1: build
     @mkdir -p benches/results
     cargo bench --bench hackers_delight
 
+# Direct SMT CLZ/CLS formula timing for issue #112; does not write JSONL.
+bench-smt-clz: build
+    cargo bench --bench smt_clz
+
 # Wipe criterion HTML reports and the JSONL accumulator.
 bench-clean:
     rm -rf target/criterion benches/results
@@ -123,6 +127,7 @@ help:
     @echo "  test          - Run tests"
     @echo "  build-tests   - Build AArch64 test binaries"
     @echo "  test-all      - Run complete test suite"
+    @echo "  bench-smt-clz - Time direct CLZ/CLS SMT equivalence queries"
     @echo "  mutants       - Run cargo-mutants locally (slow; informational)"
     @echo "  coverage      - Generate HTML coverage report via cargo-llvm-cov"
     @echo "  coverage-lcov - Generate LCOV coverage report (CI format)"
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,66 +1,99 @@
-# Plan - issue #181: Pin reversed-order `--live-out` error
+# Plan - issue #112: Replace `bv_clz` ITE chain with binary-search decomposition
 
 ## 1. Problem Restated
 
-Issue #181 asks for a narrow regression test documenting the current malformed-order behavior for `--live-out "nzcv;x0"`. The accepted grammar is `<regs>` or `<regs>;<flags>` with `nzcv` valid only in the flags half, so the reversed input currently splits into register half `nzcv` and flag half `x0`, then fails through the register parser with `invalid register name: 'nzcv'`. The PR should pin that exact diagnostic so future parser-message changes fail with an actionable test failure, without changing parser behavior.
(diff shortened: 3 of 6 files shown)


pmatos:
**Regarding the automated code review** (approved, two non-blocking suggestions): each was triaged by two independent AI reviewers — Claude Opus 4.8 and Codex GPT-5.5 — which agreed on both outcomes.

**Suggestion 2 (note why the benchmark compares each sequence to itself) — Fixed in `2099459`.**
Added a 4-line comment at the `check_equivalence_with_config_metrics` call in `benches/smt_clz.rs`: each case is compared against itself to isolate the CLZ/CLS SMT encoding/solve cost — `random_tests(0)` forces the SMT path and Z3 still builds the full `f(x) == f(x)` formula, so self-comparison stresses exactly the encoding under test rather than the cost of finding a distinct equivalent.
Validation: `cargo fmt -- --check` and `cargo clippy --bench smt_clz -- -D warnings` both pass.

**Suggestion 1 (inline comment on `chunk = upper_is_zero.ite(&lower, &upper)`) — No change.**
The `bv_clz` doc comment already states this exact rule ("If the upper half is zero … continue with the lower half; otherwise continue with the upper half"), so an inline comment at the call site would duplicate adjacent documentation and cut against the repo's "comment sparingly" guideline. Unlike the `bv_ror` precedent (a genuinely non-obvious SMTLIB2 shift edge case), this line is a faithful encoding of the already-documented algorithm. Both reviewers rejected it with high confidence.

On the other minor observations: the `BV::eq`-by-value form is intentional — it's the Clippy-preferred (`needless_borrows_for_generic_args`) spelling, so switching back to a borrow would reintroduce a lint; the `PLAN.md` and self-equivalence items were process observations, and the self-equivalence rationale is now captured by the fix above.

_Triage by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree, reply and we'll re-evaluate._

PR #380, opened 2026-06-16
fix(ir): print zero-shift movz as mov alias

### LINKED ISSUE #113: Print `MovZ { shift: 0 }` using the canonical `mov` mnemonic
Deferred from #108: `Display` for `MovZ { rd, imm, shift: 0 }` currently emits `movz x0, #imm`. In canonical AArch64 assembly the preferred form for that bit pattern is `mov x0, #imm`.

Original feedback by @claude[bot] on PR #108 (https://github.com/pmatos/s11/pull/108):

> **MovZ { shift: 0 } displays as `movz` rather than the preferred alias `mov`.**
>
> In canonical AArch64 assembly, `MOVZ Xn, #imm` with shift=0 is the preferred form `MOV Xn, #imm`. The `Display` impl emits `movz x0, #imm`, which is technically valid but non-canonical. This could confuse users reading optimizer output or when feeding the output to an assembler that checks canonical form. Consider emitting `mov` for `shift == 0` and `imm >= 0` cases (matching how `MovImm` is printed), or at minimum add a note in the IR type comment.

**Context:** `src/ir/instructions.rs` — the `MovZ` arm of the `Display` impl.

**Why deferred:** style-preference — the current form is unambiguous and round-trips through the parser (`mov` is routed to `MovImm`, `movz` is routed to `MovZ`), so collapsing `MovZ{shift:0}` to `mov` is a cosmetic / canonicality polish rather than a correctness fix. Evaluators disagreed: one DEFER (diminishing-returns), one REJECT (style-preference). Filing the lower-bound outcome so it isn't silently lost — a human can decide whether the canonical-form benefit is worth changing the round-trip property.

**Suggested approach if/when addressed:**

- Either emit `mov` for `MovZ { shift: 0 }` (and reconsider parser routing so the canonical disassembly still round-trips), or leave the `Display` impl as-is and add a comment to the variant docstring noting the deliberate divergence from the canonical AArch64 alias.

pmatos:
Summary:
- Print `Instruction::MovZ { shift: 0 }` with the canonical `mov` mnemonic.
- Keep shifted `movz` display unchanged.
- Add parser/display coverage documenting that explicit zero-shift `movz` normalizes to `mov` text and then reparses as `MovImm`.

Verification:
- `cargo test ir::instructions::tests::movz_shift0_display_uses_mov_alias -- --exact`
- `cargo test parser::tests::parse_movz_shift0_display_normalizes_to_mov_alias -- --exact`
- `cargo test parser::tests::parse_movz_and_movk_accept_canonical_forms -- --exact`
- `cargo test parser::tests::test_tier1_display_parser_roundtrip -- --exact`
- `cargo test movz`
- `cargo test --all`
- `./ci_check.sh`

Note: the requested `scripts/full_test.sh` path does not exist in this repository; `./ci_check.sh` ran the repo-present `./test_all.sh`.

Closes #113

### DIFF
--- a/src/ir/instructions.rs
+++ b/src/ir/instructions.rs
@@ -1337,7 +1337,7 @@ impl fmt::Display for Instruction {
             }
             Instruction::MovZ { rd, imm, shift } => {
                 if *shift == 0 {
-                    write!(f, "movz {}, #{}", rd, imm)
+                    write!(f, "mov {}, #{}", rd, imm)
                 } else {
                     write!(f, "movz {}, #{}, lsl #{}", rd, imm, shift)
                 }
@@ -1494,6 +1494,23 @@ mod tests {
             rm: Operand::Register(Register::X0),
         };
         assert_eq!(format!("{}", eor), "eor x0, x0, x0");
+    }
+
+    #[test]
+    fn movz_shift0_display_uses_mov_alias() {
+        let shift0 = Instruction::MovZ {
+            rd: Register::X0,
+            imm: 0x1234,
+            shift: 0,
+        };
+        assert_eq!(shift0.to_string(), "mov x0, #4660");
+
+        let shift16 = Instruction::MovZ {
+            rd: Register::X0,
+            imm: 0x1234,
+            shift: 16,
+        };
+        assert_eq!(shift16.to_string(), "movz x0, #4660, lsl #16");
     }
 
     #[test]
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2456,7 +2456,7 @@ mod tests {
     #[test]
     fn parse_movz_and_movk_accept_canonical_forms() {
         let cases = [
-            ("movz x0, #0x1234", "movz x0, #4660"),
+            ("movz x0, #0x1234", "mov x0, #4660"),
             ("movz x0, #0xFFFF, lsl #48", "movz x0, #65535, lsl #48"),
             ("movk x1, #0", "movk x1, #0"),
             ("movk x1, #0x5678, lsl #16", "movk x1, #22136, lsl #16"),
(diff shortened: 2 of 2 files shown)


pmatos:
Regarding the automated code review above — each item was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5) into FIX / DEFER / REJECT:

**Item 1 — explanatory comment on the zero-shift `MovZ` → `mov` Display branch: Deferred (diminishing returns). Tracked as #382 (https://github.com/pmatos/s11/issues/382).**
The two evaluators split (one FIX, one REJECT), so per the conservative triage rule this is deferred rather than changed in-PR. The roundtrip asymmetry the comment would describe is already named and explicitly asserted by the new `parse_movz_shift0_display_normalizes_to_mov_alias` test (`MovZ{shift:0}` → `"mov x0, #4660"` → reparses to `MovImm`), and the sibling `MovN`/`MovK` zero-shift arms set a no-comment precedent. Filed separately to avoid churn here.

**Item 2 — add a zero-shift `movz` row to `convert_capstone_op_handles_all_supported_aarch64_mnemonics`: No action taken.**
Both evaluators agree this is not needed. That test's loop only asserts each case yields `ConvertOutcome::Instruction(_)` — it asserts **no** Display text — so a `("movz", "x0, #1")` row would not "confirm the full path now produces `mov x0, #1`". It would only bump the `cases.len() == 132` tripwire to 133 while repeating `movz` coverage that is already present (`movz` is in both the cases array and the documented mnemonic set). The CLAUDE.md "extend it whenever you add an opcode" guideline isn't triggered here — this PR adds no new opcode. The `mov`-text behavior is already directly asserted in `src/ir/instructions.rs` (`movz_shift0_display_uses_mov_alias`) and `src/parser/mod.rs` (`parse_movz_shift0_display_normalizes_to_mov_alias`).

**Item 3 — decimal display of hex immediates: No action taken.** Noted; the review itself flagged this as informational with no action needed, and it is pre-existing `Display` behavior unrelated to this change.

_If you disagree with any of these, reply and we'll re-evaluate._

Issue #181, opened 2026-05-18
Add reversed-order `--live-out` parser test pinning `nzcv;x0` error message


pmatos:
Deferred from #172: add a test that exercises the reversed-order input `nzcv;x0` and asserts on the resulting error message, so a future change to the parser's error text fails loudly with an actionable failure.

Original feedback by @claude[bot] on PR #172 (https://github.com/pmatos/s11/pull/172#issuecomment-4475747275):

> `nzcv;x0` hits the `semicolon_count == 1` path, splits to `regs_part = "nzcv"` and `flags_part = "x0"`. `LiveOutRegisters::from_str("nzcv")` would fail with "invalid register", which is correct, but the error might surprise a user who wrote the clause in the wrong order. A test with an assertion on the error message would document the behavior explicitly and make the failure message more actionable if the error text ever changes.

**Context:** `src/validation/live_out.rs` test module; would add ~4 lines pinning the error message.

**Why deferred:** diminishing-returns — one evaluator preferred to fix, the other preferred to defer, on the grounds that `nzcv;x0` is already rejected through the existing register parser path and pinning that malformed-order error adds little coverage beyond the existing grammar tests. Disagreement defaults to defer per the autofix triage rules.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.7 and Codex GPT-5.4._

pmatos:
Opened PR #355.

Implementation note: the autonomous run contract referenced `scripts/full_test.sh`, but this s11 checkout does not contain that script. I chose the repository-native alternative, `./ci_check.sh`, because CLAUDE.md documents it as the pre-push CI gate and it runs formatting, build, test-binary generation, cargo tests, and `test_all.sh`. The alternative would have been to stop on the missing script, but the available repo gate covered the intended quality surface.

pmatos:
Closing as completed: the requested reversed-order `--live-out` regression test is already on `main`. PR #355 added `test_parse_live_out_contract_reversed_order_nzcv_x0_error` for `nzcv;x0`, and PR #358 later kept that exact regression while updating the pinned diagnostic to the more actionable misplaced-flag message. I verified the focused test locally with `cargo test validation::live_out::tests::test_parse_live_out_contract_reversed_order_nzcv_x0_error -- --exact` (passed). I am not opening a duplicate PR from this stale continuation branch because it is already merged upstream.

PR #383, opened 2026-06-16
refactor(aarch64): consolidate candidate opcode ids

### LINKED ISSUE #116: Consolidate duplicated `opcode_id` table between `isa/aarch64.rs` and `search/candidate.rs`
Deferred from #109: `src/search/candidate.rs::opcode_id` (free function) and `InstructionType::opcode_id` impl on `src/isa/aarch64.rs` are mirrored match tables. They are kept in lockstep by hand — every new IR variant has to be added in both places. Adding the six bit-manipulation ops in #109 required updating both, exposing the pattern again.

Original feedback by @claude on PR #109 (https://github.com/pmatos/s11/pull/109#issuecomment-4457916285, point #6):

> `src/search/candidate.rs::opcode_id` and the `InstructionType::opcode_id` impl in `src/isa/aarch64.rs` mirror each other exactly. The PR correctly adds the new cases to both. This is pre-existing technical debt, but since the PR touches both anyway, it would be a good time to remove the free function from `candidate.rs` and call `instr.opcode_id()` at the one call site that uses it.

**Why deferred:** legitimate tech debt, but pre-dates #109 and is orthogonal to "add six new IR variants". Consolidating means changing the one consumer of the free function to call through the trait, then deleting the function — small but touches unrelated code paths. Belongs in its own cleanup PR so it can be reviewed without bit-manip noise.

**Suggested fix:** replace the lone call to the free `candidate::opcode_id` with `instr.opcode_id()` via the `InstructionType` trait import, delete the free function, and confirm no other consumers exist (grep `opcode_id` in `src/search/`).

pmatos:
Summary:
- Removed the duplicated `candidate::opcode_id` AArch64 match table.
- Routed candidate coverage helpers/tests through `InstructionType::opcode_id()` from the ISA backend.
- Deleted the now-tautological candidate-vs-ISA opcode drift test.

Validation:
- `cargo test candidate::` (includes red/green cycle after deleting the helper)
- `cargo test isa::aarch64::tests::all_instruction_families_cover_trait_methods`
- `cargo test --all` (initially failed because fixture binaries were absent; passed after `./build_tests.sh`)
- `./ci_check.sh`

Closes #116

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,66 +1,65 @@
-# Plan - issue #181: Pin reversed-order `--live-out` error
+# Plan: Consolidate AArch64 opcode IDs
 
 ## 1. Problem Restated
 
-Issue #181 asks for a narrow regression test documenting the current malformed-order behavior for `--live-out "nzcv;x0"`. The accepted grammar is `<regs>` or `<regs>;<flags>` with `nzcv` valid only in the flags half, so the reversed input currently splits into register half `nzcv` and flag half `x0`, then fails through the register parser with `invalid register name: 'nzcv'`. The PR should pin that exact diagnostic so future parser-message changes fail with an actionable test failure, without changing parser behavior.
+`src/search/candidate.rs` defines a free `opcode_id(&Instruction) -> u8` match table that mirrors the canonical `InstructionType::opcode_id` implementation for `Instruction` in `src/isa/aarch64.rs`. This hand-synchronised duplicate makes each new AArch64 IR variant a two-site update and required an extra drift test. The implementation should remove the search-layer copy and route every `candidate.rs` opcode-ID use through the existing `InstructionType` trait method.
 
 ## 2. Files To Touch
 
-- `src/validation/live_out.rs` - add one unit test in the existing `#[cfg(test)] mod tests`, near the `parse_live_out_contract` malformed-input tests around lines 677-727.
-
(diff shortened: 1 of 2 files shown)


pmatos:
**Regarding the automated code review** — thanks for the approval and the three suggestions. Each was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5); outcomes below.

**1. Opcode-ID collision between sign-extension and bitfield families — _Out of scope for this PR_.**
Agreed it's worth tracking. The collision is pre-existing — this PR only consolidates a duplicated table that already encoded IDs 49–53 — and fixing it changes stochastic mutation opcode selection, which is behavioral work that belongs in its own change with dedicated tests. Tracked as #390 (https://github.com/pmatos/s11/issues/390).

**2. Mention the PLAN.md rewrite in the PR description — _No action taken_.**
This concerns PR-description text rather than source code, so there is no code change. For the record: the PLAN.md rewrite (issue #181 plan → issue #116 plan) is intentional — it is the working plan for this PR's task.

**3. Explicit scoped `use crate::isa::InstructionType;` in the test module — _Style preference (no change)_.**
The test module already accesses `InstructionType` via `use super::*` (candidate.rs:1069), which re-exports the top-level import at candidate.rs:5. It compiles today and `use super::*` is the idiomatic test-module convention; a redundant scoped import only hedges against a hypothetical future import cleanup, so no change.

PR #394, opened 2026-06-16
docs(semantics): clarify SMULH product bound

### LINKED ISSUE #125: Tighten SMULH `wrapping_mul` justification comment in concrete.rs
Deferred from #122: the comment justifying `wrapping_mul` for SMULH's i128 product is technically correct but slightly misleading — it singles out `i64::MIN × i64::MIN` as a special case when in fact every `i64 × i64` product fits in `i128` without wrapping.

Original feedback by @claude[bot] on PR #122 (https://github.com/pmatos/s11/pull/122#issuecomment-4459471773):

> The argument would be cleaner as: "the signed product of two 64-bit values always fits in 127 bits (the max magnitude is 2^63 × 2^63 = 2^126 < 2^127 − 1 = i128::MAX), so wrapping_mul never wraps." The current wording is slightly confusing because it implies only the `MIN×MIN` case is special, when actually all cases are safe.

**Context:** `src/semantics/concrete.rs` SMULH arm. Current comment:
```rust
// wrapping_mul for i128 is sufficient: a×b fits in 128 bits when
// |a|,|b| ≤ i64::MAX, except for i64::MIN×i64::MIN which still
// fits in i128 since i64::MIN as i128 is -2^63.
```

**Why deferred:** ambiguous — one evaluator favored the rewrite (clearer general argument), the other favored the original (explicit edge-case mention is load-bearing for readers verifying the boundary). The behavior is correct; this is a wording refinement only.

**Suggested resolution:** either rewrite to "the signed product of two 64-bit values fits in 127 bits (max magnitude 2^126 < i128::MAX)" or keep current wording and add a sentence noting the bound explicitly. Pick whichever is more useful when reviewing.

pmatos:
Summary:
- Clarify the SMULH concrete-interpreter comment so it states the general i64-by-i64 product bound.
- Keep the existing wrapping_mul implementation unchanged and refresh PLAN.md for issue #125.

Validation:
- cargo test semantics::concrete::tests::test_smulh_matches_rust_i128 -- --exact
- cargo fmt --all
- ./build_tests.sh
- cargo test --all (after building test binaries)
- ./ci_check.sh

Note: scripts/full_test.sh is not present in this s11 checkout; ./ci_check.sh is the repo-local full gate and passed.

Closes #125


pmatos:
Implementation note: I chose the bound-based rewrite rather than keeping the old edge-case wording plus another sentence, because the issue feedback was specifically that the old wording over-emphasized `i64::MIN * i64::MIN`. I did not add a new test because this is a comment-only clarification; the existing SMULH boundary regression was run before and after the edit. The requested `scripts/full_test.sh` gate is not present in this s11 checkout, so I used the repo-local full gate `./ci_check.sh`, which passed.

pmatos:
Regarding the automated code review on this PR — specifically the single optional observation suggesting the alternative wording *"well within i128's range of ±2^127"*:

**Style preference (no change)** — The review itself flags this as "a style preference, not a correctness issue," and the current comment is already technically correct: the signed product of two i64 values has max magnitude 2^63 × 2^63 = 2^126, which is below `i128::MAX`, so `wrapping_mul` cannot wrap. Rewording an already-accurate comment — which was the entire point of this PR — would add cosmetic churn without value. No code change made.

Issue #127, opened 2026-05-15
Complete the MNEG ↔ MADD/MSUB mutation graph


pmatos:
Deferred from #122: after the MNEG ↔ MUL bridge in commit `8950f28` (PR #122), the multiply-accumulate cluster is connected but asymmetric — `MADD` / `MSUB` mutate to `MNEG` (collapsing `ra`), but the reverse transitions don't exist. MNEG can only escape via MUL.

Original feedback by @claude[bot] on PR #122 (https://github.com/pmatos/s11/pull/122#issuecomment-4459762981):

> Asymmetry in the opcode-swap graph for MNEG:
>
>     MADD ←→ MSUB ←→ MNEG ←→ MUL
>                      ↑
>                (no path back to MADD/MSUB)
>
> `MNEG` only mutates to `MUL`; there is no `MNEG → MADD` or `MNEG → MSUB` path. The reverse edges from `MADD`/`MSUB` to `MNEG` exist, but the walk is not symmetric. This won't cause incorrect results, but the stochastic search may be slower to discover `MADD`/`MSUB` candidates when starting from a `MNEG`-containing sequence.

**Context:** `src/search/stochastic/mutation.rs::mutate_opcode`. The Mneg arm currently:
```rust
Instruction::Mneg { rd, rn, rm } => match rng.random_range(0..2) {
    0 => Instruction::Mul { rd, rn, rm },
    _ => Instruction::Mneg { rd, rn, rm },
},
```

**Why deferred:** diminishing-returns — the reviewer themselves says "won't cause incorrect results" and "Probably fine for v1." MCMC can still reach MADD/MSUB from MNEG via the `generate_random_instruction` reset strategy (already updated in #122). Search quality could improve, but the fix breaks the established `mutate_opcode` invariant (all transitions preserve register fields — adding MNEG → MADD/MSUB requires a fresh `ra` pick from `self.registers`).

**Suggested fix:** Either widen MNEG to a 4-way pick that includes MADD/MSUB with a random `ra`, or move the cluster-completion logic out of `mutate_opcode` into a dedicated multiply-family mutation strategy that's allowed to introduce new register references.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.7 and Codex GPT-5.x._

pmatos:
[...]

PR #395, opened 2026-06-16
fix(search): complete MNEG opcode mutation graph

pmatos:
Completes the stochastic opcode mutation graph for the AArch64 multiply-accumulate family by adding direct MNEG -> MADD and MNEG -> MSUB transitions. The new accumulator register is selected through the existing mutator register pool, and the mutation docs/tests cover the deliberate operand-structure exception.\n\nVerification:\n- cargo fmt --all\n- cargo clippy --all -- -D warnings\n- ./build_tests.sh\n- cargo test --all\n- ./ci_check.sh\n\nCloses #127\n\n**Merge with `gh pr merge --merge` (merge commit, not squash) per CLAUDE.md.**


pmatos:
Implementation note: I chose the minimal MNEG arm widening over a new multiply-family mutation strategy. That keeps the change local to the existing opcode-swap table, preserves rd/rn/rm exactly, and uses the existing Mutator::random_register fallback for the newly introduced ra. A dedicated strategy would give this family a cleaner abstraction boundary, but it would be broader than needed for completing the missing MNEG -> MADD/MSUB graph edges in #127.

pmatos:
Thanks for the thorough review (claude[bot]). Each point was triaged with two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5); outcomes below.

1. **Probability-distribution asymmetry** — **Deferred (diminishing returns).** Agreed this is not a correctness issue (candidates are always validated by equivalence checking); it only mildly skews the proposal distribution. Captured as a secondary item in the follow-up rather than churning this scoped PR. Tracked as #410.

2. **`Mul` has no direct edge to `Madd`/`Msub`** — **Out of scope for this PR.** Correct, and a real graph-completeness gap, but separate from completing the `MNEG → MADD/MSUB` edges (#127). Filed as a follow-up. Tracked as #410.

3. **`PLAN.md` committed as a tracked file** — **No change.** Tracking `PLAN.md` is a deliberate, established workflow in this repo, so no action (as noted in the review).

4. **Rust pattern-binding nit** — **No change.** As observed, Rust enforces identical bindings across the or-pattern; the test additionally asserts exact preserved registers, so there is nothing to fix.

5. **"Ready to merge"** — acknowledged, thank you.

Unrelated to this PR: the red `coverage` check is a **pre-existing flaky test** (`fast_only_flags_only_contract_detects_ccmp_initial_nzcv_divergence`, ~0.7% intermittent false `Equivalent` from an unseeded RNG in `fast_path_initial_nzcv_variants`). This PR touches neither `equivalence.rs` nor the RNG. Tracked separately as #409; re-running the job to clear it.

PR #402, opened 2026-06-16
fix(ir): report shifted shift-slot registers defensively

### LINKED ISSUE #132: Add defensive ShiftedRegister arm to source_registers for Lsl/Lsr/Asr/Ror shift slot
Deferred from #129: belt-and-suspenders correctness improvement.

> The shift-slot instructions (`Lsl`, `Lsr`, `Asr`, `Ror`) were not changed in `source_registers`. If someone constructs `Lsl { shift: Operand::ShiftedRegister { reg: X5, .. } }` programmatically, `source_registers` will omit X5. Since `is_encodable_aarch64` returns `false` for this case, it should never reach live-out tracking in practice. But a defensive arm here would make the contract explicit.

**Context:** `src/ir/instructions.rs::source_registers` — Lsl/Lsr/Asr arm at the existing block, plus the Ror arm. Add `Operand::ShiftedRegister { reg, .. } => regs.push(*reg)`.

**Why deferred:** ambiguous (evaluators split: one DEFER as belt-and-suspenders, one REJECT as redundant given the encodability gate). The encodability check at `is_encodable_aarch64` already rejects `ShiftedRegister` in the shift slot of these four instructions, so any code path that reasons about live-in registers will never see the bad construction. Filing as a follow-up so the defensive arm can be added (or the contract documented) without churning #129.

pmatos:
## Summary
- Add defensive `source_registers` handling for `ShiftedRegister` operands in LSL/LSR/ASR/ROR shift slots.
- Add a regression test that keeps those shapes unencodable while reporting the inner shifted register as a source.

## Tests
- `cargo test ir::instructions::tests::test_source_registers_shift_slot_shifted_register_defensive -- --exact`
- `cargo test ir::instructions`
- `cargo fmt --all`
- `./build_tests.sh`
- `cargo test --all`
- `./test_all.sh`
- `./ci_check.sh`

Closes #132

### DIFF
--- a/src/ir/instructions.rs
+++ b/src/ir/instructions.rs
@@ -1052,7 +1052,8 @@ impl Instruction {
                 match shift {
                     Operand::Register(r) => regs.push(*r),
                     Operand::ShiftedRegister { reg, .. } => regs.push(*reg),
-                    Operand::Immediate(_) | Operand::ExtendedRegister { .. } => {}
+                    Operand::ExtendedRegister { reg, .. } => regs.push(*reg),
+                    Operand::Immediate(_) => {}
                 }
                 regs
             }
@@ -1124,7 +1125,8 @@ impl Instruction {
                 match shift {
                     Operand::Register(r) => regs.push(*r),
                     Operand::ShiftedRegister { reg, .. } => regs.push(*reg),
-                    Operand::Immediate(_) | Operand::ExtendedRegister { .. } => {}
+                    Operand::ExtendedRegister { reg, .. } => regs.push(*reg),
+                    Operand::Immediate(_) => {}
                 }
                 regs
             }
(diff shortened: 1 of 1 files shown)


pmatos:
Regarding the automated code-review comment on this PR (the `ExtendedRegister` shift-slot point):

**Fixed in `62f8893`.**

Both points are addressed:

- **Main (ExtendedRegister asymmetry):** Correct — the LSL/LSR/ASR and ROR arms of `source_registers()` reported the inner `ShiftedRegister` reg but left `ExtendedRegister` in the no-op arm, while the binary-op `rm` arm and the Cmp/Cmn/Tst arm already push `ExtendedRegister.reg`. The shift-slot arms now do the same: `Operand::ExtendedRegister { reg, .. } => regs.push(*reg)`. This is squarely within "report shifted shift-slot registers defensively" (both are unencodable, programmatically-constructed IR shapes), so it's folded into this PR rather than deferred.
- **Nit (test comment):** The regression test now exercises both unencodable variants (`ShiftedRegister` *and* `ExtendedRegister`) in a nested loop, and the comment mentions both.

Changed: `src/ir/instructions.rs` — split the no-op arm in the LSL/LSR/ASR and ROR cases of `source_registers()`; extended `test_source_registers_shift_slot_shifted_register_defensive`.
Validation: `./ci_check.sh` (fmt + clippy + build + full test suite incl. the regression test) all pass.

_This outcome was reached after independent triage by two AI reviewers (Claude Opus 4.8 and Codex GPT-5.5), both of which agreed the change belongs in this PR._

pmatos:
Regarding the second automated review (the un-defensified `Ccmp`/`Ccmn` and `Bic`/`Bics`/`Orn`/`Eon`/`Adds`/`Subs`/`Ands` `rm` slots):

**Out of scope for this PR** — the observation is factually correct (lines 1088 and 1115 still use `if let Operand::Register(r) = rm`), but those are different instruction arms than the LSL/LSR/ASR/ROR shift slots this PR (issue #132) targets. The affected operand forms are themselves currently unencodable / out-of-scope (CCMP/CCMN take only Register/Immediate `rm`; ADDS/SUBS `ShiftedRegister` forms are out of scope per #59), so this is a defensive-only consistency sweep rather than an active bug — and, as the review itself notes, it's "not a blocker" and "worth opening a follow-up issue." Applying it here would expand a focused fix into an open-ended uniform sweep.

Tracked as #411 (https://github.com/pmatos/s11/issues/411).

_This outcome was reached after independent triage by two AI reviewers (Claude Opus 4.8 and Codex GPT-5.5), both of which agreed it should be deferred to a separate issue._

PR #404, opened 2026-06-16
fix(search): prefilter sp shifted-register candidates

### LINKED ISSUE #133: Pre-filter SP from shifted-register candidate enumeration
Deferred from #129: candidate-enumeration micro-optimization.

> In `generate_all_instructions`, the inner loop pushes shifted-register instructions for all registers including SP, and relies on `generate_all_encodable_instructions` to call `is_encodable_aarch64` as a second pass. This works, but if the register pool is large the wasted allocation adds up. A pre-check `if rm != Register::SP && rd != Register::SP && rn != Register::SP` inside the loop would trim those early. Not blocking, just a potential future optimization.

**Context:** `src/search/candidate.rs::generate_all_instructions` — the new shifted-register inner loop introduced by #129.

**Why deferred:** scope-creep — the post-hoc filtering is correct (just wasteful). Worth doing once we have evidence the allocation pressure matters; the existing `Operand::Register` enumeration in the same file does the same thing for SP/XZR, so this should be a coordinated cleanup not a one-off.

pmatos:
Summary:
- Update the issue #133 implementation plan in tracked PLAN.md.
- Skip SP-bearing rd/rn/rm tuples before allocating AArch64 shifted-register Add/Sub/And/Orr/Eor candidates.
- Add a regression test that raw candidate enumeration keeps valid shifted-register candidates while excluding SP from those slots.

Verification:
- cargo test test_generate_all_instructions_shifted_register_prefilters_sp
- cargo test test_generate_all_instructions_contains_shifted_register_add
- cargo test test_generate_all_instructions_includes_all_shifted_kinds_for_logical
- cargo test test_generate_all_instructions_arith_excludes_ror
- cargo test search::candidate
- ./build_tests.sh
- cargo test --all
- cargo fmt --all -- --check
- ./ci_check.sh

Closes #133


pmatos:
@claude[bot] Regarding your code review — thanks, both points are addressed:

**Ask A — `AddW`/`SubW` test gap → Fixed in `b812249`.**
Changed: extended `test_generate_all_instructions_shifted_register_prefilters_sp` to add `AddW` and `SubW` match arms, so SP-exclusion is now asserted for them too (they're generated inside the same SP-guarded block — note this branch was just merged with `main`, which introduced the 32-bit `AddW`/`SubW` shifted-register variants), instead of falling through the catch-all `_` arm unverified.
Validation: `cargo test test_generate_all_instructions_shifted_register_prefilters_sp`, `cargo clippy --all -- -D warnings`, and `./ci_check.sh` all pass.

**Ask B — exact count assertion → Deferred (diminishing returns).** Tracked as #436 (https://github.com/pmatos/s11/issues/436).
Two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5) agreed this is a low-value nit: an exact-count assertion would be louder on a regression but is more brittle to unrelated candidate-matrix changes, and `> 0` is a fine minimum sanity check (as you noted). Filed for separate follow-up rather than churning this PR.

pmatos:
Thanks for the follow-up review and the "ready to merge". Both non-blocking nits are triaged (each independently assessed by two AI reviewers, Claude Opus 4.8 and Codex GPT-5.5):

**Suggestion 1 — hoist `rd`/`rn` SP checks above the `rm` loop → Deferred (diminishing returns).** Tracked as #439 (https://github.com/pmatos/s11/issues/439).
Both reviewers also flagged that the literal snippet is unsafe as written: wrapping the whole `for &rm in registers` loop in `if rd != SP && rn != SP` would *also* gate the issue-#60 extended-register block, which must stay **outside** the SP guard (extended-register ADD/SUB permit SP in rd/rn). The perf win is negligible at current register-list sizes and a correct hoist needs careful restructuring, so this is a follow-up rather than in-PR churn. The caveat is recorded in the issue.

**Suggestion 2 — move `use crate::ir::ShiftKind;` out of the loop body → Style preference (no change).**
This `use` pre-exists this PR (it is not part of the #404 diff), function-local `use` is valid idiomatic Rust, and the nit is explicitly non-blocking with no project-backed rule requiring the move.

_These assessments were made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree, please reply and we'll re-evaluate._

PR #407, opened 2026-06-16
test(aarch64): cover ccmp ccmn follow-up cases

### LINKED ISSUE #137: CCMP/CCMN follow-up tests and sample coverage
Deferred from #128: three small CCMP/CCMN test-coverage gaps surfaced by the post-merge code review.

Original feedback by @claude on PR #128 (https://github.com/pmatos/s11/pull/128#issuecomment-4460759524 and https://github.com/pmatos/s11/pull/128#issuecomment-4460831173):

> One gap: there are no unit tests that exercise find_first_difference with flags_live=true directly (verifying the XZR sentinel and packed flag values).

> CCMP_NZCV_SAMPLES missing 8 (N-only, 0b1000): Several conditions (MI, PL, GE, LT, GT, LE) depend on N. Not a bug for a bounded sample but worth considering.

> Parser test for AL/NV as condition string in ccmp/ccmn (rejected at parse time or only by is_encodable_aarch64?).

Concrete asks:
1. Add a targeted unit test for `find_first_difference` with `flags_live=true` and divergent flags, asserting the `(Register::XZR, packed_f1, packed_f2)` sentinel shape.
2. Expand `CCMP_NZCV_SAMPLES` in `src/search/candidate.rs` from `[0, 1, 7, 15]` to include `8` (the N-only literal), or document why the current sample is sufficient.
3. Add a parser test confirming `ccmp x1, x2, #0, al` and `ccmp x1, x2, #0, nv` round-trip through `parse_line`; document whether rejection should be at parse time or only at `is_encodable_aarch64`.

**Why deferred:** diminishing-returns — the end-to-end behaviour of all three is exercised through other tests (the issue-57 acceptance, integration-path coverage of `check_equivalence_with_config` with `flags_live=true`, encoder rejection of AL/NV via `is_encodable_aarch64`). These would harden the contract surface, not fix a known bug.

pmatos:
Summary:
- Add direct flag-live `find_first_difference` coverage for the XZR/NZCV sentinel shape.
- Include the N-only `8` literal in both CCMP/CCMN NZCV sample tables and pin generation coverage.
- Document the parser boundary for CCMP/CCMN `al`/`nv`: parsed as conditions, rejected by final AArch64 encodability.

Verification:
- `cargo fmt --all`
- `cargo test --all`
- `./ci_check.sh`

Closes #137


pmatos:
Regarding the automated review summary (the two "Minor Observations"):

**Style preference (no change)** — both observations were triaged and neither warrants a code change in this PR.

**Ask A — loosen the hardcoded `cond: MI`/`PL` arms to wildcards:** Acknowledged as a reasonable readability preference, but no change. `generate_all_instructions` emits a `Ccmp`/`Ccmn` for every (nzcv-sample × condition) combination, so the `MI`/`PL` arms are present and stable as long as those conditions exist — the test is not actually iteration-order-dependent. The current form reliably pins the real invariant (`nzcv=8` is generated for both the register-`rm` `Ccmp` and immediate-`rm` `Ccmn` shapes), and loosening it would be cosmetic churn the review itself flagged as "not a blocker."

**Ask B — check for a third NZCV-sampling site (already handled):** Verified by grep. There are exactly two fixed sample tables — `src/search/candidate.rs` and `src/isa/aarch64.rs` — both already updated with `8`. The stochastic mutation path samples `nzcv` via `(rng.random::<u32>() & 0x0F)` over the full `0..=15` range, so it already covers `8` with no table to update; the symbolic path is solver-driven with no sample table. No third site exists.

_This assessment was made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.x). If you disagree, please reply and we'll re-evaluate._

PR #414, opened 2026-06-16
fix(aarch64): normalize capstone aliases before IR ingestion

### LINKED ISSUE #141: Normalize Capstone AArch64 aliases (mov-wide, cinc/cinv/cneg) before IR ingestion
Deferred from #140: Capstone emits AArch64 aliases that `parser::parse_line` cannot ingest cleanly, so `convert_to_ir` warn-and-drops them and `optimize_elf_binary` reassembles a shorter sequence — silently changing program semantics.

Original feedback by @chatgpt-codex-connector on PR #140 (https://github.com/pmatos/s11/pull/140#discussion_r3250250457):

> **Normalize Capstone aliases before skipping parser errors**
>
> When Capstone disassembles move-wide encodings such as `movz x0, #1, lsl #16`, it can emit the alias `mov x0, #0x10000`; feeding that string directly to `parse_line` constructs `MovImm` and then rejects it as not encodable, so this path returns `Unsupported` and `optimize_elf_binary` reassembles a sequence with the original instruction omitted/padded out. The same raw-alias problem applies to conditional-select aliases like `cinc`/`cinv`/`cneg`; please translate Capstone aliases to the IR forms or treat operand/encoding parse failures as fatal rather than skippable.

**Context:** `src/main.rs::convert_capstone_op` and the `ConvertOutcome::Unsupported` arm of `convert_to_ir`.

**Verified failure modes:**
- `mov x0, #0x10000` (Capstone's alias for `movz x0, #1, lsl #16`) — `parse_line` accepts it as `MovImm { imm: 0x10000 }`, then `is_encodable_aarch64` rejects because `MovImm` is limited to 16-bit immediates (`MovImm` requires `imm <= 0xFFFF`). Result: `ParseLineError::Other` → `Unsupported` → dropped.
- `cinc x0, x1, eq` / `cinv x0, x1, lt` / `cneg x0, x1, ge` — these conditional-select aliases are not in `parse_line`'s mnemonic match, so they hit `UnknownInstruction` → `Unsupported` → dropped.

**Suggested approaches (pick one or both):**
1. Add a Capstone-alias normalization layer before `parse_line`: translate `mov rd, #large_imm` → the matching `movz`/`movn`/`movk` sequence, and translate `cinc`/`cinv`/`cneg` → their `csinc`/`csinv`/`csneg` canonical forms.
2. Promote operand/encoding parse failures (`ParseLineError::Other` and out-of-range encodings) inside `convert_to_ir` from warn-and-skip to fatal — at least when the binary contains code we're being asked to optimize — so silent semantic drift can never happen.

**Why deferred:** scope-creep — PR #140 is intentionally narrow (route Capstone disassembly through the existing parser for parity with the asm-text path). Alias normalization or a fatal-on-skip policy is a substantive new feature with its own design choices, test surface, and behavioral implications for `optimize_elf_binary`. The evaluators split on whether to fix in this PR; conservative resolution is to land #140 as-is and address alias handling here.

pmatos:
## Summary
- Normalize Capstone move-wide `mov Xd, #imm` aliases to single-instruction `movz`/`movn` canonical parser text when representable.
- Normalize Capstone `cinc`/`cinv`/`cneg` aliases to `csinc`/`csinv`/`csneg` with inverted conditions.
- Keep unrepresentable aliases fatal for optimization and document the narrow Capstone-only bridge.

## Verification
- `cargo fmt --all`
- `cargo test convert_capstone_op`
- `cargo test --test integration_tests docs_capability`
- `./build_tests.sh`
- `cargo test --all`
- `./ci_check.sh`

Note: the generic run contract mentioned `scripts/full_test.sh`, but this s11 checkout has no such script; `./ci_check.sh` ran the repository `test_all.sh` gate instead.

Closes #141


pmatos:
Autonomous run notes:

- Kept Capstone alias normalization one-to-one with existing IR instructions. `mov Xd, #imm` is accepted only when it maps to one `movz` or one `movn`; constants requiring `movk` materialization still make optimization reject the selected window.
- Kept `cinc`/`cinv`/`cneg` as Capstone bridge aliases only, not parser mnemonics; docs and tests pin that distinction.
- The run contract mentioned `scripts/full_test.sh`, but this s11 checkout does not have that script. I ran the repo gates instead: `cargo fmt --all`, clippy with `-D warnings`, `cargo test --all` after `./build_tests.sh`, and `./ci_check.sh` (which runs `test_all.sh`).

pmatos:
Regarding the automated code review above — each of the five items was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5), and the disputed facts were verified by running the tests and reading `is_encodable_aarch64`. Outcomes below.

**Item 2 (test `mov x0, #5` → Unsupported) — corrected & fixed in `ee8cc01`.**
The premise that "`mov` is not a parser mnemonic" is incorrect: `parser::parse_line` dispatches `"mov" => parse_mov(...)` (src/parser/mod.rs:1749), and `parse_mov` returns `MovImm` for immediates. Verified behavior:
- `mov x0, #5` → `Instruction(MovImm { X0, 5 })` — encodable (`MovImm` allows `0..=0xFFFF`), **not** `Unsupported`.
- `mov x0, #0x12345678` → `MovImm` → fails `is_encodable_aarch64` → already pinned as rejected by the existing `convert_capstone_op_for_optimization_rejects_unnormalizable_mov_alias` test (via the `Other` arm).

So I pinned the *actual* fall-through behavior rather than the (incorrect) `Unsupported` assertion.

**Item 3 (W-register fall-through) — corrected & fixed in `ee8cc01`.**
`mov w0, #0x10000` does **not** become `Unsupported`: `parse_mov` routes W-register destinations to a logical-immediate `Orr` (src/parser/mod.rs:369-383), so the result is `Instruction(Orr { .., W32 })`. The new test `convert_capstone_op_passes_mov_alias_fall_through_to_parser` pins both `mov x0, #5` → `MovImm` and `mov w0, #0x10000` → `Orr`, locking the normalizer's fall-through boundary.

**Item 5 (UnknownInstruction diagnostic parity) — fixed in `ee8cc01`.**
Both parser-failure arms of `convert_capstone_op` now go through a shared `describe_unsupported_line` helper, so the `UnknownInstruction` arm surfaces the normalized form (`raw normalized as \`norm\``) just like the `Other` arm. A bridge regression that normalizes to an unknown mnemonic is now visible in the warning.

**Item 4 (unnecessary `.clone()`) — resolved by the Item 5 change.**
The flagged `raw_line.clone()` lived in the `UnknownInstruction` arm; that arm was rewritten to pass `&raw_line` into the shared helper, so the redundant clone is gone. (Note: `raw_line` is still used by the sibling `Other` arm, so it could not simply be moved before this refactor — the helper resolves both cleanly.)

**Item 1 (dead `imm != 0` guard) — deferred (diminishing returns).**
Both reviewers agreed this is a cosmetic dead-code clarification not worth code churn in this PR. Tracked as #438 (https://github.com/pmatos/s11/issues/438).

Validation: `cargo test convert_capstone_op` (10 passing) and the full `./ci_check.sh` gate both pass.

pmatos:
Regarding the second automated code review (posted after the `origin/main` merge) — triaged again by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Outcomes:

**Item 2 (missing AL/NV negative test) — fixed in `89bdfd0`.**
Both reviewers agreed this is a genuine, untested error path. Added `convert_capstone_op_rejects_cond_select_al_nv_aliases`, which pins `cinc x0, x1, al` and `cinv x2, x3, nv` → `Unsupported` (also covering `convert_capstone_op`'s normalize-error → `Unsupported` arm, previously unexercised). Validated by the full `./ci_check.sh` gate.

**Item 1 (variable shadowing in `normalize_mov_wide_alias`) — deferred.**
Pure readability nit on pre-existing code with no behavioral impact. Tracked as #441 (https://github.com/pmatos/s11/issues/441).

**Item 3 (`raw_line.clone()` on the `Ok(None)` branch) — no change.**
Agreed with the review's own note: `raw_line` is reused later by `describe_unsupported_line` (and the `Err` arm), so the clone is structurally required, not a defect.

**Item 4 (test comment length) — no change.**
The comment documents the non-obvious normalization boundary (why `mov x0,#5` → `MovImm` and `mov w0,#0x10000` → `Orr` both bypass the move-wide normalizer) — exactly the kind of WHY CLAUDE.md wants documented. Trimming it would drop load-bearing context.

One note on the **`imm != 0` guard** in `move_wide_movz_encoding`: this review calls it "necessary defensive code," whereas the first review called it "dead code, remove it." The contradiction is itself the reason that item was deferred rather than churned (#438) — the guard stays.

CI is green and the branch is now up to date with `main`.

pmatos:
Regarding the third automated code review — all three items are informational/already-tracked; no code change is warranted, so this converges here.

**Item 1 (dead `imm != 0` guard) — already tracked (#438), no change.**
This is now the third contradictory verdict on the same lines: round 1 said "dead code, remove it," round 2 said "necessary defensive code, the right move," and this round says "effectively dead code, misleading." That instability is precisely why it was deferred rather than churned. The guard stays; the clarify-or-remove discussion lives in #438 (https://github.com/pmatos/s11/issues/438).

**Item 2 (`mov x0, #0` test) — no change (same path already pinned).**
`mov x0, #0` exercises the identical code path as the already-pinned `mov x0, #5`: `0..=0xffff` → `normalize_mov_wide_alias` returns `Ok(None)` → `parse_mov` → `MovImm` (encodable). It is the same equivalence class, not a distinct branch, so `convert_capstone_op_passes_mov_alias_fall_through_to_parser` already locks the behavior. (Note: `value = 0` never reaches `move_wide_movz_encoding`, so this does not exercise the item-1 guard either.)

**Item 3 (diagnostic-path asymmetry) — no change.**
As the review notes, this is correct: the normalize-error arm (`format!("{raw_line} ({err})")`) and the parser-error arm (`describe_unsupported_line`) are intentionally distinct — the former reports a Capstone alias the bridge itself rejected, the latter reports a parser failure after normalization succeeded. Both already carry the raw spelling. Noted for the future if the bridge grows.

All substantive feedback from the earlier rounds is addressed; CI is green and the branch is even with `main`. Closing the loop here.

PR #415, opened 2026-06-16
feat(aarch64): derive search flag liveness from context

### LINKED ISSUE #138: Derive flags_live from analysis instead of hardcoding with_flags(true) in search
Deferred from #128. Both AArch64 optimization callers now hardcode `EquivalenceConfig::with_flags(true)`:

- `src/search/stochastic/mcmc.rs:206` (MCMC verification step)
- `src/search/symbolic/synthesis.rs:238` (symbolic synthesis verification step)

This was required to keep the softened `flag_writers_diverge` sound: with `flags_live=false`, the solver no longer sees the divergence the old structural guard caught. The conservative default rejects any rewrite that disturbs NZCV across the optimization window, even when flags are demonstrably dead after the window.

Original feedback by @claude on PR #128 (https://github.com/pmatos/s11/pull/128#issuecomment-4460931872):

> `with_flags(true)` is hard-wired in stochastic and symbolic search (mcmc.rs, synthesis.rs). This is sound but over-constrains the optimizer: any rewrite that touches flags — even when flags are demonstrably dead after the window — will be rejected. In practice, sequences containing `CMP`/`ADDS`/`SUBS` followed by consumers that only care about register values (e.g. `MADD`, `MUL`, `LSL`) will hit a wall.

**Suggested fix:** introduce a flag-liveness analysis (mirror of the existing `LiveOutRegisters`) that scans the surrounding context to decide whether NZCV is observed after the optimization window. Plumb the result into `EquivalenceConfig::with_flags(...)` instead of unconditionally `true`. Useful precedent: `LiveOutRegisters` in `src/semantics/live_out.rs`.

**Why deferred:** scope-creep — this is a real follow-up but architectural and outside the scope of "add CCMP/CCMN". The current hardcoded `true` is sound; this issue tracks unlocking the CCMP-based rewrites the acceptance test demonstrates (where the two forms have intentionally different final NZCV).

pmatos:
Summary
- derive AArch64 ELF optimization flags_live from fixed terminators and known fall-through suffixes
- make AArch64 enumerative, stochastic, and symbolic backend verification honor live_out.flags_live()
- add regression coverage for flags-dead CMP removal and conservative flags-live rejection
- update ADR notes for the new search flag-liveness behavior

Conservative choices
- non-flag fixed terminators keep NZCV live because this PR does not build a CFG or prove successor paths
- empty suffixes, decode failures, unsupported instructions, and unanalysed control-flow edges keep NZCV live
- LLM search remains conservative per ADR-0002

Verification
- cargo fmt --all -- --check
- ./build_tests.sh
- cargo test --all
- just test
- ./ci_check.sh

Closes #138


pmatos:
Thanks for the review. The two optional / non-blocking suggestions were triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5):

**1. Comment on the `ConvertOutcome::Skip` arm — Style preference (no change).** The empty `Skip` arm is already clear from the surrounding match (modifies-flags → dead, terminator → live, unsupported → conservative) and its NOP pass-through is pinned by the `convert_capstone_op_skips_nop_silently` test. The repo guideline is to comment sparingly (only complex code), so this would be a cosmetic addition. No code change.

**2. Test for `optimization_context_for_backend` non-AArch64 path — Deferred (diminishing returns).** A legitimate coverage gap, but exercising the `arch != Aarch64` branch needs a non-trivial x86 `ElfPatcher` + `TextSection` + `Capstone` fixture the suite currently lacks — disproportionate to pinning a one-line `default()` return on an already-approved, fully-green PR. Tracked as #429 (https://github.com/pmatos/s11/issues/429).

The other items in the review (the `RET` conservatism, the `flags_read_before_overwrite_after_window` alias, and the section-boundary note) were all explicitly flagged as intentional / fine-as-is, so no action there.

PR #420, opened 2026-06-16
fix(search): bound non-sp random operand selection

### LINKED ISSUE #147: pick_non_sp in random instruction generators can loop forever in release builds
Deferred from #143: harden `pick_non_sp` closure against pathological all-SP register sets.

Original feedback by @claude-review on PR #143 (https://github.com/pmatos/s11/pull/143):

> The debug_assert! fires in debug mode, but in release builds a pathological register set consisting only of SP would loop forever. The existing pattern elsewhere (e.g., the enumerative generator) uses a continue guard before entering the register loop, which is safer.

**Context:** the same `pick_non_sp` closure pattern exists in two places in `src/search/candidate.rs`:

- Lines 488–514 — CCMP/CCMN block, already on `main` (introduced before #143).
- Lines 520–545 — bit-field block, introduced in #143.

Both rely on `debug_assert!(registers.iter().any(|r| *r != Register::SP))` to detect the degenerate input, and both compile that assertion out in release. With `registers = [SP]` the inner `loop { ... if r != SP { break } }` never terminates.

**Why deferred:** scope-creep — fixing only the bit-field site leaves the same hazard in CCMP/CCMN; fixing both is a wider refactor not specific to #143. The default register sets used by the CLI (`src/main.rs`) and search config (`src/search/config.rs`) explicitly exclude SP from the random pool, so the bug is unreachable in normal operation.

**Suggested fix:** at the top of each random-instruction arm (or in a shared helper at the function entry), filter SP from a local view of `registers` once and use that filtered slice instead of `pick_reg`. Or convert the closure to a bounded retry that returns an error / falls back to a different opcode after N attempts.

pmatos:
## Summary
- filter SP out of the CCMP/CCMN and bit-field random candidate pools before operand sampling
- fall back finitely to the existing MNEG-shaped random candidate pattern when a pathological pool contains only SP
- add budgeted RNG regressions for forced CCMP/CCMN and bit-field slots with SP-only and mixed SP/non-SP pools

## Decision Notes
I used the existing finite fallback pattern already present in `src/isa/aarch64.rs` instead of adding retry limits or changing opcode probabilities. This keeps the random slot mapping unchanged while removing the release-build infinite-loop path.

## Verification
- `cargo test generate_random_instruction_` (red before the fix, green after)
- `cargo test search::candidate::tests`
- `./build_tests.sh`
- `cargo test --all`
- `./test_all.sh`
- `./ci_check.sh`

Note: the requested `scripts/full_test.sh` path does not exist in this repository; `test_all.sh` and `ci_check.sh` are the repo-local equivalents.

Closes #147

### DIFF
--- a/src/search/candidate.rs
+++ b/src/search/candidate.rs
@@ -1156,6 +1156,9 @@ mod tests {
         }
     }
 
+    // Inverts `random_range`'s Lemire mapping: returns the smallest 32-bit
+    // word `w` such that `(w * range) >> 32 == value`, so a `BudgetedRng` can
+    // be primed to force a specific `random_range(0..range)` outcome.
     fn word_for_range(range: u32, value: u32) -> u32 {
         assert!(range > 0);
         assert!(value < range);
@@ -1165,6 +1168,11 @@ mod tests {
         word
     }
 
+    // Primes a `BudgetedRng` to steer `generate_random_instruction` down a
+    // chosen opcode arm: the first draw is the `rd` index (`random_range(0..2)`
+    // for these 2-register pools), the second is the opcode slot
+    // (`random_range(0..38)`). Slot 32 = CCMP/CCMN arm, slot 33 = bit-field arm
+    // — keep these in sync with the match in `generate_random_instruction`.
     fn rng_for_opcode_slot(slot: u32, rd_index: u32) -> BudgetedRng {
         BudgetedRng::new(vec![word_for_range(2, rd_index), word_for_range(38, slot)])
     }
--- a/src/search/candidate.rs
+++ b/src/search/candidate.rs
@@ -812,19 +812,21 @@ pub fn generate_random_instruction<R: rand::RngExt>(
             // (encoded in the Xn slot, not XSP). `generate_all_instructions`
             // filters SP at enumeration time; mirror that here so the
             // mutator does not bleed avoidable is_encodable_aarch64
-            // rejections. The debug_assert guards against a degenerate
-            // `registers = [SP]` caller, which would otherwise spin
-            // forever in the retry loop below.
-            debug_assert!(
-                registers.iter().any(|r| *r != Register::SP),
-                "CCMP/CCMN random generator requires at least one non-SP register",
-            );
-            let pick_non_sp = |rng: &mut R| loop {
-                let r = pick_reg(rng);
-                if r != Register::SP {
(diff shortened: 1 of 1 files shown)


pmatos:
Autonomous decision note: I used the existing finite MNEG fallback pattern from src/isa/aarch64.rs for SP-only CCMP/CCMN and bit-field random slots, rather than adding retry limits or changing opcode probabilities. I also treated the missing scripts/full_test.sh as a repo-contract mismatch and ran the repo's test_all.sh plus ci_check.sh instead.

pmatos:
Re: the automated code review on this PR — each suggestion was triaged with two independent AI reviewers (Claude Opus 4.8 + Codex GPT-5.5). Outcomes:

**Fixed in `fe3dcd0`:**
- **`word_for_range` comment** — added a comment noting it inverts `random_range`'s Lemire mapping (smallest `w` such that `(w * range) >> 32 == value`).
- **Test slot-number coupling** — `rng_for_opcode_slot` now documents that the two primed words are `(rd index, opcode slot)` and that slot 32 = CCMP/CCMN arm, slot 33 = bit-field arm, with a note to keep them in sync with the match in `generate_random_instruction`.

Validation: `./ci_check.sh` (fmt + build + `build_tests.sh` + `cargo test` + `test_all.sh`) — all green.

**Deferred (diminishing returns)** — *comment on the MNEG fallback*: slot 32 already carries a 6-line comment covering the SP-filter rationale and the finite fallback, so the reviewers split FIX/DEFER on whether the extra one-liner (specifically *why `Mneg(SP, SP, SP)` is safe given `is_encodable_aarch64` rejection*) is worth changing an otherwise-green PR. Tracked as #432 (https://github.com/pmatos/s11/issues/432).

**No change taken** — *extract the duplicated `non_sp`/MNEG block into a shared helper*: both reviewers rejected this. The 12-line block is duplicated across only two adjacent match arms; the early `return` plus the borrow-capturing `pick_non_sp` closure make a clean extraction awkward; and the cited "three similar lines > premature abstraction" CLAUDE.md principle does not exist in the repository — the project's documented stance (ADR-0004) actually leans *against* premature abstraction.

_The DEFER/REJECT assessments above were made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree, reply and we'll re-evaluate._

PR #423, opened 2026-06-16
test(search): guard candidate family sampling balance

### LINKED ISSUE #153: Normalise per-slot sampling probability in candidate.rs::generate_random_instruction
Deferred from #143: rebalance the random-instruction slot table so each opcode family has equal sampling weight.

Original feedback by @claude[bot] on PR #143 (https://github.com/pmatos/s11/pull/143#issuecomment-4463356716):

> `match rng.random_range(0..31)` — 31 slots (0–30), explicit arm `28 =>` for bit-field (1/31 ≈ 3.2%), `_ =>` now catches two values (29 and 30), giving multiply-accumulate 2/31 ≈ 6.5%. This doubles the sampling probability of multiply-accumulate instructions relative to bit-field.

**Context:** `src/search/candidate.rs::generate_random_instruction` uses a numeric-slot match against `rng.random_range(0..N)` to dispatch into opcode families. The current pre-PR structure already had a `_` catch-all (originally covering 2 of 30 slots for multiply-accumulate, ≈ 6.7%). PR #143 added bit-field as an explicit slot 28 and bumped the range to `0..31`, so:

| Family | Pre-PR weight | Post-PR weight |
|---|---|---|
| Most families (explicit slots) | 1/30 ≈ 3.3% | 1/31 ≈ 3.2% |
| Multiply-accumulate (catch-all) | 2/30 ≈ 6.7% | 2/31 ≈ 6.5% |
| Bit-field | n/a | 1/31 ≈ 3.2% |

The bot's framing ("multiply-accumulate probability doubled") is technically incorrect — its absolute weight slightly *decreased*. But the relative asymmetry between multiply-accumulate (2x) and bit-field (1x) is real and inherits from a pre-PR design choice.

**Why deferred:** scope-creep — the asymmetry isn't a correctness issue and pre-dates this PR (multiply-accumulate has had 2x weight relative to all other explicit slots since it was added in `dd9a21c`). Equalising slot weights is a one-line refactor (move multiply-accumulate to its own explicit slot, change `_ => ...` to `unreachable!()`) but it affects every existing instruction-family's sampling weight slightly and deserves a clean review pass on its own. The sister generator `AArch64InstructionGenerator::generate_random` in `src/isa/aarch64.rs` already uses the cleaner explicit-slot pattern with `_ => unreachable!()` and is internally consistent — worth aligning `candidate.rs` to match.

**Suggested follow-up fix** in `src/search/candidate.rs::generate_random_instruction`:
- Move multiply-accumulate from the `_` catch-all to explicit slot 29.
- Change `rng.random_range(0..31)` back to `0..30`.
- Replace `_ => ...` with `_ => unreachable!()` to prevent silent drift on future additions.

pmatos:
Summary
- Add deterministic regression coverage for candidate.rs random family sampling balance.
- Assert bit-field and multiply-accumulate families are sampled at comparable rates through generate_random_instruction.
- Leave the current explicit 38-slot implementation intact because the older 31-slot issue text is stale.

Closes #153

### DIFF
--- a/src/search/candidate.rs
+++ b/src/search/candidate.rs
@@ -1422,7 +1422,15 @@ mod tests {
         let mut rng = ChaCha8Rng::seed_from_u64(0x153);
         let mut bitfield_count = 0u32;
         let mut madd_count = 0u32;
-        const DRAWS: u32 = 76_000;
+
+        // Both families occupy exactly one top-level slot in the
+        // `rng.random_range(0..SLOTS)` dispatch, so "equal sampling weight" means
+        // equal probability of *entering* the slot (~1/SLOTS), not equal per-variant
+        // weight: slot 33 makes a secondary 0..6 draw (plus SP-rejection retries) and
+        // slot 34 a 0..5 draw. DRAWS is sized so each family is expected EXPECTED_PER_FAMILY times.
+        const SLOTS: u32 = 38;
+        const EXPECTED_PER_FAMILY: u32 = 2_000;
+        const DRAWS: u32 = SLOTS * EXPECTED_PER_FAMILY;
 
         for _ in 0..DRAWS {
             let instr = generate_random_instruction(&mut rng, &regs, &imms);
@@ -1442,12 +1450,29 @@ mod tests {
             }
         }
 
+        // Each count is ~Binomial(DRAWS, 1/SLOTS); the std-dev of their difference is
+        // sqrt(2 * DRAWS * (1/38) * (37/38)) ≈ 63, so a tolerance of 250 is ≈ 4σ
+        // (p_fail ≈ 1e-5) — tight enough to catch a slot-weight regression, loose
+        // enough not to flake.
         let delta = bitfield_count.abs_diff(madd_count);
         assert!(
             delta <= 250,
             "bit-field and multiply-accumulate should have equal top-level sampling weight \
              over {DRAWS} draws, got bitfield={bitfield_count}, madd={madd_count}, delta={delta}",
         );
+
+        // Delta alone would still pass if both families collapsed to the same wrong
+        // rate (e.g. ~500 each), so bound each absolute count near the expected value
+        // to also catch a degenerate dispatch.
+        for (name, count) in [
+            ("bit-field", bitfield_count),
+            ("multiply-accumulate", madd_count),
+        ] {
+            assert!(
+                (1_500..=2_500).contains(&count),
+                "{name} family count {count} is far from the expected {EXPECTED_PER_FAMILY} over {DRAWS} draws",
+            );
+        }
     }
 
     #[test]
--- a/src/search/candidate.rs
+++ b/src/search/candidate.rs
@@ -1413,6 +1413,44 @@ mod tests {
     }
 
     #[test]
+    fn generate_random_instruction_samples_bitfield_and_madd_families_evenly() {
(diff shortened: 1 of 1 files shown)


pmatos:
Thanks for the review. Each suggestion was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5), which agreed on every item.

**Fixed in `86a3a3c`.**

- **(1) Threshold rationale** — added a comment deriving the `<= 250` tolerance as ≈4σ (σ of the difference of two Binomial(DRAWS, 1/38) ≈ 63, p_fail ≈ 1e-5).
- **(2) Slot vs per-variant weight** — added a comment clarifying that "equal sampling weight" means equal probability of *entering* the top-level slot (~1/38), not equal per-variant weight, calling out slot 33's secondary `0..6` draw (+ SP-rejection retries) vs slot 34's `0..5` draw.
- **(3) `DRAWS` derivation** — `DRAWS` is now `SLOTS * EXPECTED_PER_FAMILY` (`38 * 2_000`), tying it to the dispatch arity and the expected per-family count.
- **(5) Absolute-count guard** — added assertions bounding each family's count to `[1_500, 2_500]`, so a degenerate dispatch that under-samples both families equally (delta still ≤ 250) is now caught as well.

Changed: `src/search/candidate.rs` — test `generate_random_instruction_samples_bitfield_and_madd_families_evenly` (comments, derived constants, two absolute-count assertions).
Validation: targeted `cargo test` for this test passes; full `./ci_check.sh` (fmt + build + unit tests + test-binary builds + full suite) is green.

**Style preference (no change) — (4) assertion-message capitalisation.** Left as-is: you marked it non-blocking, it changes no behaviour, and cosmetic-only churn needlessly re-triggers review on an otherwise-green PR.

_This triage was made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree, reply and we'll re-evaluate._

pmatos:
Thanks for the second pass — and for confirming the round-1 statistics and the absolute-count guard. The three new suggestions were triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). No code change is pushed here (none of the three is a correctness issue, and re-pushing only re-triggers another review round); outcomes below.

**(1) `SLOTS = 38` duplicates the production `0..38` literal — Out of scope for this PR.** Exposing a `pub(crate)` slot-count constant (or otherwise tying the test to the dispatch) is a production-code maintainability change beyond this test-only PR. The test already comments the linkage, and the new `[1_500, 2_500]` per-family bounds fail loudly if the slot count drifts far, so there's no present correctness risk. Tracked as #437 (https://github.com/pmatos/s11/issues/437).

**(2) Add a comment clarifying the 2-of-38 scope — No action taken.** The narrow scope is already evident from the test name (`..._samples_bitfield_and_madd_families_evenly`), the existing slot-vs-variant comment, and the `Closes #153` reference; an extra comment adds churn without clarity gain.

**(3) Hoist the repeated `rand`/`rand_chacha` imports — No action taken.** As noted, this is a pre-existing pattern not introduced by this PR and explicitly non-blocking; touching sibling tests would expand scope for zero correctness value.

_This triage was made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree, reply and we'll re-evaluate._

PR #426, opened 2026-06-16
feat(stochastic): produce extended-register mutations

### LINKED ISSUE #151: Stochastic mutation: produce and bridge ExtendedRegister operands
Deferred from #144: the `Operand::ExtendedRegister` variant is mutation-frozen — `mutate_operand` strips it to a plain register / immediate, `mutate_opcode` never proposes the form, and the standalone SXTB/SXTH/SXTW/UXTB/UXTH instructions are self-bridging stubs in `src/search/stochastic/mutation.rs`. The enumerative search and the SMT path discover the form; the MCMC path cannot.

Original feedback by @claude on PR #144 (https://github.com/pmatos/s11/pull/144#issuecomment-4462932632):

> `mutate_operand` folds `ExtendedRegister` into the same arm as `Register`/`ShiftedRegister`, mutating it to a plain register or immediate with probability 1. The stochastic engine can never *propose* an ExtendedRegister — only evaluate ones inherited from the initial candidate pool. The `mutate_opcode` branches for the standalone extends are also identity-only.

**Context:** `src/search/stochastic/mutation.rs`. The follow-up needs (mirroring the shifted-register precedent from #59):

1. `random_extended_register<R: RngExt>(&self, rng: &mut R) -> Operand` — pick a random non-SP/non-XZR register, random `ExtendKind`, random shift in `0..=4`.
2. Wire it into the `mutate_operand` arms for ADD/SUB/CMP/CMN with a probability split alongside `random_shifted_register`.
3. `mutate_opcode` bridges among SXTB/SXTH/SXTW/UXTB/UXTH/CLZ/CLS/RBIT/REV/REV32/REV16 (currently the five extends self-loop).
4. A `strip_extended_for_logical` analogue of `strip_ror_for_arith` so opcode mutations into AND/ORR/EOR/TST don't carry an invalid operand variant.

**Why deferred:** scope-creep — explicitly carved out of #144's scope ("Stochastic mutation: the variant is mutation-frozen for now (no `random_extended_register` helper, no opcode bridges). Adding it is a polish slice; the issue's acceptance is met without it." in the PR description). The MCMC search still finds the form via the cross-pollination from the enumerative candidate pool.

_Filed automatically by `pm-autofix-pr` after triage by Claude Opus 4.7._

pmatos:
Summary:
- add stochastic sampling for AArch64 ExtendedRegister operands on ADD/SUB/CMP/CMN
- strip ExtendedRegister operands when opcode mutation bridges into logical/TST forms
- bridge standalone extend opcodes into the CLZ/CLS/RBIT/REV single-source peer cluster

Verification:
- cargo fmt --all -- --check
- cargo test search::stochastic::mutation
- cargo test enumerate_emits_add_with_extended_register
- cargo test enumerate_emits_cmp_with_extended_register
- ./build_tests.sh
- cargo test --all
- ./ci_check.sh

Closes #151

### DIFF
--- a/src/search/stochastic/mutation.rs
+++ b/src/search/stochastic/mutation.rs
@@ -16,7 +16,7 @@
 
 use crate::ir::instructions::MOVW_LEGAL_SHIFTS;
 use crate::ir::types::Condition;
-use crate::ir::{Instruction, Operand, Register, RegisterWidth};
+use crate::ir::{ExtendKind, Instruction, Operand, Register, RegisterWidth};
 use crate::search::candidate::generate_random_instruction;
 use crate::search::config::MutationWeights;
 use rand::RngExt;
@@ -37,6 +37,34 @@ fn strip_ror_for_arith(rm: Operand) -> Operand {
         Operand::Register(reg)
     } else {
         rm
+    }
+}
+
+/// Drop the extended-register modifier when bridging from arithmetic
+/// opcodes (ADD/SUB/CMP/CMN — extended form allowed) to logical opcodes
+/// (AND/ORR/EOR/TST — extended form rejected). Other operand shapes pass
+/// through unchanged.
+fn strip_extended_for_logical(rm: Operand) -> Operand {
+    if let Operand::ExtendedRegister { reg, .. } = rm {
+        Operand::Register(reg)
+    } else {
+        rm
+    }
+}
+
+fn single_source_opcode_peer<R: RngExt>(rng: &mut R, rd: Register, rn: Register) -> Instruction {
+    match rng.random_range(0..11) {
+        0 => Instruction::Clz { rd, rn },
+        1 => Instruction::Cls { rd, rn },
+        2 => Instruction::Rbit { rd, rn },
+        3 => Instruction::Rev { rd, rn },
+        4 => Instruction::Rev32 { rd, rn },
+        5 => Instruction::Rev16 { rd, rn },
+        6 => Instruction::Sxtb { rd, rn },
+        7 => Instruction::Sxth { rd, rn },
+        8 => Instruction::Sxtw { rd, rn },
+        9 => Instruction::Uxtb { rd, rn },
+        _ => Instruction::Uxth { rd, rn },
     }
 }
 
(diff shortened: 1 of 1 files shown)


pmatos:
Implementation notes for the autonomous run:

- I did not update `docs/spec/*.md` because this s11 workspace has no `docs/spec/` directory and the change is internal to AArch64 stochastic mutation, not syntax/CLI/semantic surface.
- The generic contract command `scripts/full_test.sh` is not present here. I ran the repo-local gate instead: `./ci_check.sh`, which builds fixtures, runs `cargo test`, and runs `./test_all.sh`.
- `cargo test --all` initially failed only because `binaries/` fixtures were absent in the reused workspace; after `./build_tests.sh`, `cargo test --all` passed.

pmatos:
Regarding the automated code review above (comment 4719340192) — each suggestion was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Outcomes:

**1. Double-iteration in `random_extended_register` — Deferred (diminishing returns).** `self.registers` is ≤16 elements and the suggested `Vec` trades two cheap iterations for a heap allocation in a hot path with no correctness gain. Tracked as #434.

**2. Compile-time cluster-size assertion — No action taken (not an issue).** `SINGLE_SOURCE_OPCODE_CLUSTER` is typed `[SingleSourceOpcode; 11]`, so `assert!(…len() == 11)` is tautological — its length is fixed by the type and the assert can never fire. It does **not** catch drift between the production `random_range(0..11)` / match-arm count and the cluster; that coupling is already guarded by the runtime reachability test `test_single_source_opcode_mutation_reaches_extend_and_bitmanip_peers` (asserts all 11 peers reachable and no unexpected peer). The proposed assert would add dead code that catches nothing.

**3. Document the sampling-band contingency — Deferred (diminishing returns).** The existing doc-comment already says it "falls back to the plain register/immediate distribution"; an extra note on probability-mass redistribution is marginal under the project's "comment sparingly" rule. Tracked as #434.

**4. Document the `allow_ror` invariant — No action taken (not an issue).** `allow_ror` only feeds `random_shifted_register`; `ExtendedRegister` uses `ExtendKind`, which has no ROR variant, so there is no interaction to document. A comment here would describe a non-problem.

**5. `test_` prefix consistency — Deferred.** Rust test discovery doesn't require the prefix and the module already has mixed naming, so the reviewers split FIX/REJECT; deferred rather than churn the PR. Tracked as #434.

PR #446, opened 2026-06-16
fix(search): check timeout inside length-two enumeration

### LINKED ISSUE #162: Add inner-loop timeout check in enumerative `run_length_two`
Deferred from #154: in `run_length_two`, the **timeout** check only fires at the outer (`instr1`) loop level. With a 10k-candidate pool, a worker mid-way through its inner loop can keep grinding for up to ~1 s of pre-filter rejections before the outer-level timeout in *another* worker sets `stop` and the inner-loop's stop check picks it up.

Original feedback by @claude[bot] on PR #154 ([comment](https://github.com/pmatos/s11/pull/154#issuecomment-4466293926)):

> In `run_length_two`, the timeout check only runs at the **outer** (`instr1`) level. With a large instruction pool (N ~ thousands), a single outer iteration executes N inner iterations without a timeout check. Under a tight timeout (e.g., the 50 ms test) this can cause significant overshoot on the first outer iteration. The inner loop in `run_length_two` should check `shared.stop.load(Ordering::Relaxed)` periodically too — the test even acknowledges this with a 5 s wall-clock upper bound for a 50 ms timeout.

**Context:** `src/search/enumerative/search.rs` inner loop of `run_length_two` (lines ~169-189). The inner loop already polls `shared.stop` every iteration; the missing piece is the inner-level `EnumerativeSearch::timed_out(start, config.timeout)` check that turns `stop` true *from inside* the inner loop. Adding it costs one `Instant::now()` per inner iteration (~tens of ns) but tightens worst-case timeout overshoot from ~1 s to negligible.

**Why deferred:** diminishing-returns — the existing 5 s slack in `search_honors_timeout` already absorbs the worst case, and the trade-off between hot-path `Instant::now()` cost and timeout precision is worth a focused decision (e.g. check timeout once every K inner iterations) rather than bundling into an unrelated PR.

pmatos:
Adds an inner-loop timeout check to run_length_two so a worker can set the shared stop flag while walking instr2 candidates, before allocating/costing the next candidate. Adds a regression test that forces the deadline to expire during the first inner candidate and verifies the second candidate is not costed.

Verification:
- cargo fmt --all -- --check
- cargo test search::enumerative::search::tests::run_length_two_sets_stop_when_inner_deadline_expires
- cargo test search::enumerative::search::tests
- cargo test --all
- ./ci_check.sh

Closes #162

### DIFF
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -361,13 +361,17 @@ fn run_length_two<I>(
         if shared.stop.load(Ordering::Relaxed) {
             return;
         }
-        // Mirror symbolic's timeout granularity (instr1 level).
+        // Let idle workers stop before claiming a new outer-loop item.
         if EnumerativeSearch::<I>::timed_out(start, config.timeout) {
             shared.stop.store(true, Ordering::Relaxed);
             return;
         }
         for instr2 in all_instructions {
             if shared.stop.load(Ordering::Relaxed) {
+                return;
+            }
+            if EnumerativeSearch::<I>::timed_out(start, config.timeout) {
+                shared.stop.store(true, Ordering::Relaxed);
                 return;
             }
             let mut candidate = vec![*instr1, *instr2];
@@ -518,7 +522,10 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::sync::atomic::AtomicUsize;
+
     use crate::ir::{Instruction, Operand, Register};
+    use crate::isa::{ISAMutator, InstructionType, U64};
     use crate::search::config::SymbolicConfig;
 
     #[test]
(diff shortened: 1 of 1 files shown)


pmatos:
**Deferred (diminishing returns)** — Triage of the two non-blocking notes in the automated code-review comment above, by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5):

- **Global static counter coupling** (`INNER_TIMEOUT_COST_CALLS`): legitimate but inert today — only one test references the counter, so there is no cross-test interference. A thread-local / closure-captured counter is a future cleanup, not worth churning this approved, all-green PR.
- **Test timing margin (25 ms / 50 ms):** the 2x margin is fine as-is; widening it is conditional on an *observed* CI flake, and none has occurred.

Both are tracked for later as #454. No code change in this PR.

_If you'd prefer either addressed now rather than deferred, reply and we'll re-evaluate._

pmatos:
**Already handled (no change)** — Re: the Codecov patch-coverage report above. The `codecov/patch` check run is green (`success`); the repo treats patch coverage as informational, so it does not gate merge. The flagged "50 missing lines" are predominantly test-only trait-impl scaffolding (`InnerTimeoutIsa` / `InnerTimeoutInstruction` / `InnerTimeoutMutator` methods required to satisfy trait bounds but not all invoked by the single targeted test). The production change — the inner-loop timeout guard — is directly exercised by the new regression test. Adding calls solely to raise the percentage would be test theater with no behavioral value.

_This assessment was made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5)._

PR #447, opened 2026-06-16
perf(search): cache enumerative candidate pool

### LINKED ISSUE #163: Cache enumerative candidate pool across `search()` calls
Deferred from #154: `generate_all_encodable_instructions(&config.available_registers, &config.available_immediates)` is called fresh on every `EnumerativeSearch::search()` invocation. For the current single-call CLI use this is irrelevant — but a batch consumer (e.g. per-basic-block invocation) would pay the allocation on every call.

Original feedback by @claude[bot] on PR #154 ([comment](https://github.com/pmatos/s11/pull/154#issuecomment-4466293926)):

> `generate_all_encodable_instructions` is called fresh on every invocation. For the current CLI use-case (one call per invocation) this is fine, but if `EnumerativeSearch` is ever embedded in a loop (e.g., called per basic block), the per-call allocation will matter. A `with_config` pre-generation pattern (similar to `SymbolicSearch`) would future-proof this.

**Context:** `src/search/enumerative/search.rs` early in `SearchAlgorithm::search`. Sibling [issue #160](https://github.com/pmatos/s11/issues/160) covers caching the rayon `ThreadPool`; this issue covers caching the candidate pool. The two could land together.

**Why deferred:** diminishing-returns — no batch consumer exists today; adding `cores`/`registers`/`immediates`-change-detection complexity for a benefit only future callers would observe is not worth the API churn in this PR.

pmatos:
Summary:
- cache the enumerative single-instruction candidate pool on each EnumerativeSearch instance
- add with_config(&SearchConfig) for eager candidate generation
- cover reuse, invalidation, reset preservation, and AArch64 regression behavior

Verification:
- cargo fmt --all -- --check
- cargo test --all
- ./ci_check.sh

Closes #163

### DIFF
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -38,6 +38,12 @@ struct SharedState<I: ISA> {
     best: Mutex<Option<Vec<I::Instruction>>>,
 }
 
+struct CandidatePool<I: ISA> {
+    registers: Vec<I::Register>,
+    immediates: Vec<i64>,
+    instructions: Vec<I::Instruction>,
+}
+
 impl<I: ISA> SharedState<I> {
     fn new(initial_best_cost: u64) -> Self {
         Self {
@@ -293,6 +299,7 @@ where
 
 pub struct EnumerativeSearch<I: ISA = AArch64> {
     statistics: SearchStatistics,
+    candidate_pool: Option<CandidatePool<I>>,
     _isa: PhantomData<I>,
 }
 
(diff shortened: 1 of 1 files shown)


pmatos:
Thanks for the review (`claude[bot]`). Each point was triaged with two independent evaluators (Claude Opus 4.8 + Codex gpt-5.5); outcomes below.

**3. Missing register-change invalidation test — ✅ Fixed in `d3c35fb`.**
Added `regenerates_candidate_pool_when_registers_change` to `src/search/enumerative/search.rs`. The mock ISA's `registers_from_config` is constant, so the test exercises the register branch on AArch64 (where the register pool varies with config): it warms the cache with a one-register config, re-queries with a two-register config on the same instance, and asserts the pool regenerates — a stale pool (register change ignored) would compare equal. Validation: `cargo test`, `cargo fmt --check`, and `./ci_check.sh` all pass locally.

**2. Ordered-vector cache key — ❌ No action taken (not an issue).**
As the review itself notes, this isn't worth a code change unless the infrastructure produces different orderings in practice. It doesn't: `registers_from_config` / `immediates_from_config` derive their vectors deterministically in a fixed order from the same `SearchConfig`, so `[0,1]` vs `[1,0]` can't arise for one config. The order-sensitivity is intentional and already pinned by `regenerates_candidate_pool_when_effective_immediates_change`. Normalizing would add cost for a case that cannot occur.

**1 & 4. `reset()` and `with_config` intent comments — 📋 Deferred → #461.**
Both are comment-only nits. The preservation contract is already enforced by `reset_preserves_candidate_pool`, and `with_config`'s doc comment plus the `_for_config` method name already convey the eager-population intent. Against the project's "comment sparingly, only comment complex code" convention these are churn-only in this PR, so they're tracked for the next change to this file: https://github.com/pmatos/s11/issues/461.

_(Unrelated to the review: `d3c35fb` also fixes the failing `coverage` job — `small_config()` now uses `with_timeout_option(None)` so the bounded AArch64 length-1 search can't be spuriously aborted by a 10s wall-clock timeout under llvm-cov instrumentation, mirroring the x86 sibling tests.)_

_Triaged by two independent AI reviewers (Claude Opus 4.8 and Codex gpt-5.5). If you disagree with any outcome, reply and we'll re-evaluate._

PR #450, opened 2026-06-16
fix(cli): avoid duplicate disasm ELF arch parse

### LINKED ISSUE #266: Disasm CLI arm reads + parses ELF twice (detect_cli_arch_from_elf + analyze_elf_binary)
Deferred from #255 (which only addressed the `s11 opt` arm's double-parse): the same double `fs::read` + `ElfBytes::minimal_parse` pattern still exists on the `s11 disasm` arm.

`src/main.rs` Disasm arm flow when `--arch` is supplied:

1. `detect_cli_arch_from_elf(&binary)` — `fs::read` + `minimal_parse` to cross-check the `--arch` value against the ELF's `e_machine`.
2. `analyze_elf_binary(&binary, true)` — another `fs::read` + `minimal_parse` to disassemble.

ELF files in this project's test corpus are small so the cost is negligible (same rationale as #88). The fix mirrors what #255 did on the Opt arm: build an `ElfPatcher` once at dispatch and thread it into `analyze_elf_binary`, or expose a thin path-based arch-only detection method.

Original feedback by @claude[bot] on PR #255 (https://github.com/pmatos/s11/pull/255#issuecomment-4478636396):

> `detect_cli_arch_from_elf` is now used only by the Disasm arm, which also calls `analyze_elf_binary` (another `fs::read`). Same double-read pattern exists there. Out of scope for this PR, but worth an issue for the follow-up.

**Context:** `src/main.rs` Disasm arm (`Commands::Disasm { binary, arch } => { ... }`).

**Why deferred:** scope-creep — the reviewer self-deferred this ("Out of scope for this PR, but worth an issue for the follow-up"). It's the natural Disasm-side counterpart to #88, but the API touched (`analyze_elf_binary` rather than the two optimization helpers) is different and warrants its own review.

pmatos:
Summary:
- move disasm --arch validation into analyze_elf_binary after its single ELF read/parse
- keep RISC-V disasm rejected before file I/O and preserve mismatch stderr shape
- add analyzer and CLI regression coverage for expected arch matching/mismatch

Verification:
- cargo fmt --all
- ./build_tests.sh
- cargo test --all
- ./ci_check.sh
- scripts/full_test.sh was requested by the run contract but is not present in this repository; ./ci_check.sh runs test_all.sh here.

Closes #266

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -255,13 +255,8 @@ enum Commands {
 
 // --- ELF Binary Analysis ---
 
-/// Read an ELF's `e_machine` and map it to the matching `CliArch` variant.
-/// Returns an error if the binary can't be read or the architecture isn't
-/// one we support.
-fn detect_cli_arch_from_elf(path: &Path) -> Result<CliArch, Box<dyn std::error::Error>> {
-    let data = fs::read(path)?;
-    let elf = ElfBytes::<AnyEndian>::minimal_parse(&data)?;
-    match elf.ehdr.e_machine {
+fn cli_arch_from_e_machine(machine: u16) -> Result<CliArch, Box<dyn std::error::Error>> {
+    match machine {
         elf::abi::EM_AARCH64 => Ok(CliArch::Aarch64),
         elf::abi::EM_X86_64 => Ok(CliArch::X86_64),
         elf::abi::EM_386 => Ok(CliArch::X86_32),
@@ -269,7 +264,11 @@ fn detect_cli_arch_from_elf(path: &Path) -> Result<CliArch, Box<dyn std::error::
     }
 }
 
-fn analyze_elf_binary(path: &PathBuf, disasm_mode: bool) -> Result<(), Box<dyn std::error::Error>> {
+fn analyze_elf_binary(
+    path: &Path,
+    disasm_mode: bool,
+    expected_arch: Option<CliArch>,
+) -> Result<(), Box<dyn std::error::Error>> {
     if !disasm_mode {
         println!("Analyzing ELF binary: {}", path.display());
     }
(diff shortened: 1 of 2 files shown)


pmatos:
**Triage of the automated code review** — each concern was evaluated by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5).

**1. String-based error discrimination (the main concern) — Fixed in `7bf5768`.**
Changed: extracted a shared `const ARCH_MISMATCH_PREFIX: &str = "Architecture mismatch:"` and routed the analyzer's message, the disasm caller's `starts_with` check in `main()`, the opt-path mismatch message, and the unit test through it — so renaming the diagnostic can no longer silently break the discrimination. (The integration test asserts on real CLI stderr across the binary/integration-test crate boundary, so it keeps the literal by necessity.)
Validation: `./ci_check.sh` — fmt check, build, unit tests, and the full integration suite all pass.

**2. `unreachable!` RISC-V arms — Deferred.** Correct, but introducing a narrower `SupportedArch` enum is a broader refactor beyond this PR's focused goal of removing the duplicate ELF parse. Tracked as #455.

**3. `{:?}` Debug format in error messages — Deferred.** As noted in the review, this pre-existed in the replaced code (no regression here); a `Display` impl on `CliArch` is a cross-cutting change touching all call sites. Tracked as #456.

**Codecov coverage note — no action.** Patch coverage is 95% and both `codecov/patch` and `codecov/project` checks are green; the remaining uncovered lines are the defensive `unreachable!` arms.

_The two deferrals and the no-action call were each agreed by both reviewers. If you disagree with any of these, reply here and we'll re-evaluate._

pmatos:
**Triage of the second automated code review** — each item was evaluated by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5).

**1. String-prefix error discrimination → typed enum — No change.** The shared `ARCH_MISMATCH_PREFIX` const landed in `7bf5768` — that was the first review's own stated *minimal fix* — and it already carries a doc comment noting that the disasm caller depends on the prefix (the fallback ask in this re-review). Promoting `analyze_elf_binary` to a typed `DisasmError` enum is a larger refactor beyond this PR's focused goal of removing the duplicate ELF parse; both reviewers agreed it isn't warranted here.

**2. `unreachable!` RISC-V arms (`DetectedArch`) — Already tracked.** Same item as the first review's concern #2; tracked as #455.

**3. Integration-test fixture assumption (one-line comment) — Deferred.** The reviewers split (fix-now vs. leave-it), so per this PR's policy it becomes a tracked follow-up rather than another push. Tracked as #467.

**4. `{:?}` Debug-format spellings — Already tracked.** Same item as the first review's concern #3; tracked as #456.

All checks are green on `7bf5768`. Holding further code changes here to avoid re-review churn — every push re-runs the reviewer and the remaining points are captured in the linked issues.

_Each verdict was agreed by both reviewers (item 3 was conservatively deferred on disagreement). Reply here if you'd like any of these reopened._

PR #451, opened 2026-06-16
fix(x86): prune self-cmov candidates

### LINKED ISSUE #284: Filter rd==rs no-op CMOV variants from x86 candidate pool
Deferred from #268: `cmov rax, rax, *` is provably useless; filter `rd == rs` in `generate_all_x86_instructions` to shrink the candidate pool by `n*16`.

Original feedback by @claude[bot] on PR #268 (https://github.com/pmatos/s11/pull/268):

> The comment says these are "left in for symmetry" but `cmov rax, rax, E` is a provable NOP — it only modifies rax when ZF=1, and the value it writes is rax itself. Including all n² (rd,rs) pairs for all 16 conditions inflates the search space with provably useless candidates. Consider filtering `rd == rs` in `generate_all_x86_instructions` to reduce the candidate pool by n*16.

**Context:** `src/search/candidate_x86.rs::generate_all_x86_instructions` — CMOV enumeration loop (`for &rd in registers { for &rs in registers { ... } }`).

**Why deferred:** diminishing-returns — the wasteful candidates are correctness-safe (just slow); other reg-reg variants (Add/Sub/etc) also include rd==rs by symmetry; filtering CMOV-specifically would be inconsistent without a broader pool-pruning pass.

pmatos:
Summary:
- Filter rd == rs CMOV variants from the x86 generate_all candidate pool.
- Add generator regressions for self-CMOV pruning and full expanded-pool counts.

Tests:
- cargo fmt --all
- cargo test --all
- ./ci_check.sh

Note: scripts/full_test.sh is not present in this repository; ./ci_check.sh ran the available ./test_all.sh gate.

Closes #284

### DIFF
--- a/src/isa/x86.rs
+++ b/src/isa/x86.rs
@@ -1040,9 +1040,13 @@ impl InstructionGenerator<X86Instruction> for X86InstructionGenerator {
             }
         }
         // CMOVcc is rewritable and reads flags, so enumerate every
-        // condition for each register pair. Jcc remains excluded.
+        // condition for each non-identical register pair. Jcc remains
+        // excluded.
         for &rd in registers {
             for &rs in registers {
+                if rd == rs {
+                    continue;
+                }
                 for &cond in &X86Condition::ALL {
                     out.push(X86Instruction::Cmov { rd, rs, cond });
                 }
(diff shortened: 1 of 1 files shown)


pmatos:
Regarding the `claude[bot]` code review (two items raised alongside the approval):

**1. `generate_random` / `mutate` still emit self-CMOVs — Out of scope for this PR.** Confirmed and correct: `generate_random` (opcode `14 => Cmov`) and `mutate`'s source-changing strategy draw `rd`/`rs` independently and can still produce `CMOV rd, rd` on the stochastic/MCMC path. This is not a correctness issue (the concrete interpreter treats self-CMOVs as faithful no-ops and Metropolis-Hastings rejects them), and PR #451 is a targeted `generate_all` fix that closes #284, so extending the same guard to the stochastic generator/mutator belongs in a follow-up rather than expanding this PR's scope. Tracked as #453 (https://github.com/pmatos/s11/issues/453).

**2. Remove the per-function `use ...InstructionGenerator;` import in the new test — Already handled (no change).** The nit's precondition does not hold: `mod tests` declares only `use super::*;` at the top, with no module-level `InstructionGenerator` import, and all six test functions declare their own per-function import. The new `x86_generator_filters_self_cmov_candidates` test follows that exact established convention; removing its import would break compilation and diverge from the module pattern, so no change is warranted.

PR #452, opened 2026-06-16
fix(search): sample logical bitmask immediates in stochastic mutation

### LINKED ISSUE #374: stochastic search: curated bitmask-immediate sampler for W32/X64 logical mutation
Deferred from #366: stochastic mutation of W32 (and X64) logical immediates draws from the 12-bit-tuned immediate table, so most mutated logical immediates fail `is_encodable_aarch64` and burn search iterations.

Original feedback by @claude[bot] on PR #366 (https://github.com/pmatos/s11/pull/366):

> **W32 mutations produce almost-always-invalid immediates (search efficiency)** — Only a small subset of 32-bit values are valid AArch64 logical bitmask immediates, so nearly every W32 mutation produces an unencodable candidate filtered by `is_encodable_aarch64()`. This does not affect correctness, but it will significantly hurt stochastic search efficiency for W32 programs. A `random_logical_imm32` helper (or sampling from the same table filtered to 32-bit valid values) would fix this.

**Context:** `src/search/stochastic/mutation.rs` (W32 logical operand mutation; existing in-code note at lines 865-870).

**Why deferred:** diminishing-returns — the review's premise that the X64 path uses a separate pre-validated table is inaccurate (both the W32 and X64 paths sample the same `config.available_immediates` via `random_immediate`). The real issue — a 12-bit-tuned table is a poor source for bitmask immediates, which affects the X64 logical path too — is already explicitly documented and deferred in-code at `mutation.rs:865-870` ("…until a curated bitmask-immediate table lands"), related to #65. A curated logical-immediate sampler belongs with that tracked work, not this focused PR.

pmatos:
## Summary
- add curated W32/X64 AArch64 logical bitmask immediate pools for stochastic mutation
- route logical operand mutations and arithmetic-to-logical opcode bridges through the bitmask sampler
- clamp logical bitmask immediates when opcode mutation bridges back to arithmetic immediates

## Tests
- cargo fmt --all
- cargo test --all (after ./build_tests.sh populated fixture binaries)
- ./ci_check.sh

Closes #374

### DIFF
--- a/src/search/stochastic/mutation.rs
+++ b/src/search/stochastic/mutation.rs
@@ -32,7 +32,7 @@ const LOGICAL_IMM32_POOL: &[i64] = &[
     0x8000_0000,
     0x5555_5555,
     0xaaaa_aaaa,
-    -256,
+    -256, // i64 -256 = 0xFFFF_FF00 as W32: a 24-bit high run, a valid logical bitmask immediate
 ];
 const LOGICAL_IMM64_POOL: &[i64] = &[
     0x1,
@@ -46,7 +46,7 @@ const LOGICAL_IMM64_POOL: &[i64] = &[
     0xaaaa_aaaa_aaaa_aaaa_u64 as i64,
     0xf0f0_f0f0_f0f0_f0f0_u64 as i64,
     0x8000_0000_0000_0000_u64 as i64,
-    -256,
+    -256, // i64 -256 = 0xFFFF_FFFF_FFFF_FF00 as X64: a 56-bit high run, a valid logical bitmask immediate
 ];
 
 /// Drop ROR from a shifted-register operand when bridging from a logical
@@ -1121,6 +1121,9 @@ impl Mutator {
         width: RegisterWidth,
         allow_shifted: bool,
     ) -> Operand {
+        // W32-always-immediate policy: W32 logical operands are sampled as
+        // bitmask immediates only (no register/shifted operands), unlike the
+        // X64 path below which also samples registers and shifted registers.
         if width == RegisterWidth::W32 {
             return Operand::Immediate(self.random_logical_immediate(rng, width));
         }
@@ -1155,6 +1158,10 @@ impl Mutator {
             Operand::ExtendedRegister { reg, .. } if width == RegisterWidth::X64 => {
                 Operand::Register(reg)
             }
+            // W32-always-immediate policy (see `random_logical_operand`): the
+            // X64 arms above preserve register/shifted/extended forms, so this
+            // catch-all only ever sees W32 operands. They are deliberately
+            // resampled as bitmask immediates rather than kept as registers.
             Operand::Register(_)
             | Operand::ShiftedRegister { .. }
             | Operand::ExtendedRegister { .. } => {
--- a/src/search/stochastic/mutation.rs
+++ b/src/search/stochastic/mutation.rs
@@ -22,6 +22,32 @@ use crate::search::config::MutationWeights;
 use rand::RngExt;
 
 const ADDRESS_OFFSET_POOL: [i64; 8] = [0, 8, 16, 24, 32, 64, -8, -256];
+const LOGICAL_IMM32_POOL: &[i64] = &[
+    0x1,
+    0x2,
+    0x4,
+    0x8,
+    0xff,
+    0xffff,
+    0x8000_0000,
+    0x5555_5555,
+    0xaaaa_aaaa,
+    -256,
+];
+const LOGICAL_IMM64_POOL: &[i64] = &[
+    0x1,
+    0x2,
+    0x4,
+    0x8,
+    0xff,
+    0xffff,
+    0xffff_ffff,
+    0x5555_5555_5555_5555,
+    0xaaaa_aaaa_aaaa_aaaa_u64 as i64,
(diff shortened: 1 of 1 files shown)


pmatos:
Regarding your code review above: each of the five numbered items was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Both agreed on every item. Outcomes below.

---

**Item 1 — W32 logical operand mutation never samples registers — Deferred (out of scope).**
Correct observation. Adding register-operand sampling for W32 is a search-policy expansion beyond this PR's stated immediate-sampling scope (it needs its own immediate/register split tuning and updated test expectations), so it shouldn't ride along here. Tracked as #466 (https://github.com/pmatos/s11/issues/466).

**Item 2 — `logical_rm_from_existing` catch-all converts W32 registers → immediates — Fixed in `e58ea2b`.**
Added a comment on the catch-all arm, plus one at the policy's origin in `random_logical_operand`, documenting the deliberate W32-always-immediate policy so the X64/W32 asymmetry isn't a maintenance surprise.
- Changed: `src/search/stochastic/mutation.rs` — comments only, no behavior change.
- Validation: `cargo fmt -- --check`, `cargo build`, and `cargo clippy --all -- -D warnings` all clean. (Comment-only change — cannot affect test outcomes; the full `cargo test` suite was unchanged from the green PR run.)

**Item 3 — `-256` in `LOGICAL_IMM32_POOL` is opaque — Fixed in `e58ea2b`.**
Annotated both `-256` pool entries with their masked meaning: `0xFFFF_FF00` (W32, 24-bit high run) and `0xFFFF_FFFF_FFFF_FF00` (X64, 56-bit high run) — both valid logical bitmask immediates. Same commit and validation as item 2.

**Item 4 — old TST-immediate comment retired — No change (not an issue).**
Agreed the removal is correct: `TST #imm` aliases `ANDS XZR, Xn, #imm` and is encodable. The ask was only to flag the intent in prose rather than change code; the intent is now recorded in this thread. No code change.

**Item 5 — `clamp_imm12` can produce `#0` — No change (not an issue).**
Agreed: `0xff00.rem_euclid(0x1000) == 0` yields a vacuous-but-encodable `ADD/SUB Xn, Xm, #0`, which the MCMC cost/equivalence machinery already discards or scores appropriately. Adding a code comment for an inert, correct edge case would conflict with the project's sparse-comment guideline. No change.

---

_Each FIX / DEFER / REJECT outcome above was agreed by both AI reviewers. If you disagree with any, reply and we'll re-evaluate._

pmatos:
Regarding the follow-up review above — each of the five items was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). No new code changes this round; outcomes below.

---

**Item 1 — dead code path / implicit invariant in `logical_rm_from_existing` — Deferred.**
Correct that the catch-all is currently unreachable: all 10 call sites pass `RegisterWidth::X64` (the `Add`/`Sub` → `And`/`Orr`/`Eor` bridges are 64-bit; `AddW` only bridges to `SubW`, so no W32 arithmetic→logical bridge exists). However, the suggested `assert_eq!(width, RegisterWidth::X64)` would be wrong — that arm is deliberately W32-capable and encodes the documented W32-always-immediate policy, so asserting X64-only would panic the very behavior it provides. This is intrinsically part of the W32 register-sampling follow-up already tracked in #466 (resolving that issue is what makes the arm live and decides whether W32 registers are kept or resampled), so I've noted it there rather than file a duplicate. No code change.

**Item 2 — pool asymmetry between W32 and X64 — Deferred (out of scope).**
Real and intentional: the pools are curated heuristic samples, not a correctness matrix, so exact W32/X64 parity isn't required (every entry is validated encodable by the pool test). Documenting the curation criteria (and optionally adding a `0xf0f0_f0f0` / high-half W32 entry) is a sampling-policy/doc task beyond this PR's mechanical pool introduction. Tracked as #469 (https://github.com/pmatos/s11/issues/469).

**Item 3 — W32 `0xaaaa_aaaa` as positive i64 — No change (not an issue).**
Self-resolved in the review ("Both are correct as written"). `0xaaaa_aaaa` is a positive `i64` (`0x0000_0000_AAAA_AAAA`) and every W32 pool entry is validated through `is_encodable_aarch64` in the pool test. No ask, no change.

**Item 4 — `clamp_imm12` rem_euclid comment / large-bitmask test — No change (already handled).**
`clamp_imm12` already carries a multi-line doc comment explaining the imm12 range requirement (issue #87), and `v.rem_euclid(0x1000)` is self-evidently a modular reduction — an extra inline comment would conflict with the project's sparse-comment guideline. The existing bridge test already exercises a non-imm12 immediate (`0xff00`); a larger bitmask hits the identical branch and adds no coverage (`rem_euclid` is in-range by construction). No change.

**Item 5 — strict `panic!` in the bridge test — No change (not an issue).**
The test already asserts `is_encodable_aarch64` (the suggested "softer" check) alongside the `panic!`. The `panic!` deliberately pins this PR's central invariant — arithmetic-immediate sources always become logical bitmask immediates on bridge — and should fail loudly if that regresses. Loosening it would weaken a meaningful guard; the "depends on intent" framing confirms it's intentional. No change.

---

_Each outcome was agreed by both AI reviewers. If you disagree with any, reply and we'll re-evaluate._

PR #472, opened 2026-06-18
fix(search): bound w32 shifted-register mutation amounts

### LINKED ISSUE #134: Tune stochastic mutation probability for shifted-register operands
Deferred from #129: heuristic-tuning follow-up.

> `random_operand_3op` uses a 15% probability to produce a `ShiftedRegister`. The TST-specific branch in `mutate_instr` also uses 15%. For the stochastic search to reliably discover shifted-register collapses, the heat needs to be high enough that it explores this shape frequently. Worth monitoring in practice; no action needed now.

**Context:** `src/search/stochastic/mutation.rs` — `random_operand_3op` and the Tst arm in `mutate_operand`.

**Why deferred:** ambiguous (one evaluator DEFER for "tune later", one REJECT for "no failing evidence"). 15% was a defensible starting heuristic; revisit once we have benchmark data showing the stochastic search misses or finds shifted-register optimizations at a measurable rate.

pmatos:
Summary:
- make stochastic shifted-register amount sampling width-aware
- keep X64 sampling at the tuned 30% heat and preserve amount 32 exploration
- prevent W32 AddW/SubW and MovRegW -> AddW mutation paths from emitting shifted amount 32

Closes #134

### DIFF
--- a/src/search/stochastic/mutation.rs
+++ b/src/search/stochastic/mutation.rs
@@ -23,6 +23,8 @@ use rand::RngExt;
 
 const ADDRESS_OFFSET_POOL: [i64; 8] = [0, 8, 16, 24, 32, 64, -8, -256];
 const SHIFTED_REGISTER_OPERAND_PROBABILITY: f64 = 0.30;
+const SHIFTED_REGISTER_AMOUNTS_X64: [u8; 7] = [1, 2, 3, 4, 8, 16, 32];
+const SHIFTED_REGISTER_AMOUNTS_W32: [u8; 7] = [1, 2, 3, 4, 8, 16, 31];
 
 /// Drop ROR from a shifted-register operand when bridging from a logical
 /// opcode (AND/ORR/EOR/TST — ROR allowed) to an arithmetic opcode
@@ -185,10 +187,7 @@ impl Mutator {
                     *imm = self.random_immediate(rng);
                 }
             }
-            Instruction::Add { rd, rn, rm }
-            | Instruction::AddW { rd, rn, rm }
-            | Instruction::Sub { rd, rn, rm }
-            | Instruction::SubW { rd, rn, rm } => {
+            Instruction::Add { rd, rn, rm } | Instruction::Sub { rd, rn, rm } => {
                 let choice = rng.random_range(0..3);
                 match choice {
                     0 => *rd = self.random_register(rng),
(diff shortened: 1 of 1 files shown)


pmatos:
Regarding the automated Code Review above — each point was triaged by two independent AI reviewers (Claude Opus 4.8 + Codex GPT-5.5):

**1. Doc comment on the W32 amount constant — Fixed in `d9f4bc2`.**
Added a doc comment on `SHIFTED_REGISTER_AMOUNTS_W32` citing the AArch64 `0..=31` shifted-register constraint (ARM C3.5.2) and noting that amount 32 is valid only for the X-form.

**2. `SubW` probe in `w32_arithmetic_mutation_call_sites_use_w32_shift_amounts` — Fixed in `d9f4bc2`.**
Added an explicit `SubW` operand-mutation probe with its own `saw_subw_operand_shifted` coverage guard, so the shared `AddW | SubW` arm is self-documenting and protected against a future arm split.
Validation: `./ci_check.sh` (fmt + build + full test suite) passed, including the new probe's reachability assertion.

**3. W32 logical (`And`/`Orr`/`Eor`) shifted-register operands — Deferred (out of scope).**
Agreed this is a real unexplored proposal space, but it is pre-existing and unrelated to the amount-bounding fix here, and it also requires extending `is_encodable_aarch64` to accept W-form logical shifted-register operands. Tracked as #473 (https://github.com/pmatos/s11/issues/473).

**4. `AddW → SubW` opcode mutation passing `rm` through unchanged — No change.**
As noted in the review, this is a pre-existing race, not introduced by this PR. A surviving `ShiftedRegister { amount: 32 }` carried into a `SubW` is rejected by `is_encodable_aarch64` (`ir/instructions.rs`), and MCMC skips non-encodable proposals (`mcmc.rs`), so there is no observable defect. No code change and no issue.

_If you disagree with any of these, reply and we'll re-evaluate._

pmatos:
Regarding the re-run Code Review above — triaged by two independent AI reviewers (Claude Opus 4.8 + Codex GPT-5.5):

**Doc comment for `random_operand_3op`'s `width` parameter — Fixed in `34afbf4`.**
Extended the doc comment to describe the new `width: RegisterWidth` argument (selects the W32 `≤31` vs X64 `≤32` shift-amount pool, forwarded to `random_shifted_register`).
Validation: `./ci_check.sh` (fmt + build + full test suite) passed.

**`MovRegW → SubW` opcode-mutation path — Confirmed: no such path exists (no change).**
Checked `mutate_opcode`: the `MovRegW` arm constructs only a W32-guarded `AddW` (`random_operand_3op(.., RegisterWidth::W32)`) or stays `MovRegW` — there is no `MovImmW`, so it deliberately mirrors `MovReg → MovImm` with a W-form `AddW`. The only `SubW` productions are the `AddW ↔ SubW` opcode swaps, which carry `rm` through unchanged (no fresh amount sampling); a stale `amount: 32` carried that way is rejected by `is_encodable_aarch64` and skipped by MCMC. So no `MovRegW → SubW` guard or probe is needed.

_If you disagree with either, reply and we'll re-evaluate._

pmatos:
Regarding the third Code Review above — triaged by two independent AI reviewers (Claude Opus 4.8 + Codex GPT-5.5). All three points are minor/cosmetic and taken **without a code change** (the fix is green and the substantive items from the earlier rounds are resolved):

**1. `random_operand_3op` doc wording ("X64 allows 32") — Style preference (no change).**
The comment opens with "selects the shift-amount **pool**", so `32` already reads as the top of the X64 sampling pool, not an encoding ceiling. As the review notes, the phrasing "is not wrong in context", so it's left as-is.

**2. `// X64-only, no W-form` comment on the Cmp/Cmn arm — Style preference (no change).**
The arm is already correct (there is no `CmpW`/`CmnW`); the suggestion is, as noted, purely stylistic and not load-bearing.

**3. W32 pool gap 16→31 — No action (observation only).**
Intentional, and it mirrors the existing X64 pool sparsity; flagged as an observation, not a change request.

Thanks for the thorough pass. The substantive items from the earlier rounds — W32 constant doc, the `SubW` probe, the `width` doc, and the `MovRegW → SubW` confirmation — are all addressed, and #473 tracks the deferred W32-logical-operand follow-up.

_This assessment was made by two independent AI reviewers. If you'd like any of these changed anyway, reply and we'll re-evaluate._

PR #476, opened 2026-06-18
fix(cli): print arch mismatches with CLI names

### LINKED ISSUE #206: Implement Display for CliArch and use it in opt/disasm mismatch errors
Deferred from #201: the `opt --arch` mismatch error (and the pre-existing `disasm --arch` mismatch error) formats `CliArch` with `{:?}`, which prints Rust variant names (`Aarch64`, `X86_64`, `X86_32`) instead of the kebab-case strings the user typed on the command line (`aarch64`, `x86-64`, `x86-32`).

Original feedback by @claude[bot] on PR #201 (https://github.com/pmatos/s11/pull/201):

> The error messages use `{:?}` on `CliArch`, which prints `Aarch64`, `X86_64`, `X86_32` rather than the kebab-case strings the user typed (`aarch64`, `x86-64`, `x86-32`). The `disasm` path has the same issue so this PR is at least consistent, but the mismatch between what was typed and what the error echoes back is confusing.
>
> Suggested fix: implement `Display` for `CliArch` mapping each variant to its CLI string, then switch both `disasm` and `opt` error paths from `{:?}` to `{}`.

**Context:** `src/main.rs` — the `opt` branch's "Architecture mismatch" `eprintln!` and the analogous one on the `disasm` path.

**Why deferred:** scope-creep — this is a UX polish that touches the existing `disasm` path too; PR #201 is intentionally a narrow correctness fix for the `opt` pipeline, and the bot itself flagged the item as "not a blocker for this PR."

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.7 and Codex GPT-5.4-Codex._

pmatos:
Summary:
- Implement Display for CliArch using the clap value spellings.
- Use Display in disasm and opt architecture mismatch diagnostics.
- Pin mismatch stderr tests against CLI names instead of Rust enum variants.

Tests:
- cargo fmt --all
- cargo test --all
- ./ci_check.sh
- scripts/full_test.sh (not present in this repo)

Closes #206

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -118,6 +118,19 @@ pub enum CliArch {
     X86_32,
 }
 
+impl std::fmt::Display for CliArch {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let value = match self {
+            CliArch::Aarch64 => "aarch64",
+            CliArch::Riscv32 => "riscv32",
+            CliArch::Riscv64 => "riscv64",
+            CliArch::X86_64 => "x86-64",
+            CliArch::X86_32 => "x86-32",
+        };
+        f.write_str(value)
+    }
+}
+
 impl From<DetectedArch> for CliArch {
     fn from(arch: DetectedArch) -> Self {
         // DetectedArch is the closed set of architectures ElfPatcher accepts
@@ -289,7 +302,7 @@ fn analyze_elf_binary(
         && expected_arch != detected_arch
     {
         return Err(format!(
-            "{ARCH_MISMATCH_PREFIX} --arch {expected_arch:?} but ELF reports {detected_arch:?}"
+            "{ARCH_MISMATCH_PREFIX} --arch {expected_arch} but ELF reports {detected_arch}"
         )
         .into());
     }
@@ -2174,9 +2187,7 @@ fn main() {
             let cli_arch = match arch {
                 Some(a) if a == detected_arch => a,
                 Some(a) => {
-                    eprintln!(
-                        "{ARCH_MISMATCH_PREFIX} --arch {a:?} but ELF reports {detected_arch:?}"
-                    );
+                    eprintln!("{ARCH_MISMATCH_PREFIX} --arch {a} but ELF reports {detected_arch}");
                     std::process::exit(1);
                 }
                 None => detected_arch,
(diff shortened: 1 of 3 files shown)


pmatos:
Regarding the automated code review on this PR — the two asks were triaged independently by two AI reviewers (Claude Opus 4.8 and Codex GPT-5.5), who agreed on both:

**1. "Display strings can drift from clap" — Fixed in `2762c23`.**

Changed: `impl Display for CliArch` now writes `self.to_possible_value().expect("CliArch has no skipped variants").get_name()` instead of a hardcoded `match`, so the spelling is derived from clap's `ValueEnum` and stays in sync with the CLI parser by construction (a renamed variant or a `#[value(name = ...)]` attribute can no longer drift the error message from what users type).

Validation: `cargo fmt -- --check` and `cargo clippy --all -- -D warnings` are clean, and `cli_arch_display_uses_cli_value_names` + `analyze_elf_binary_rejects_expected_arch_mismatch` pass — confirming clap's `get_name()` yields exactly `aarch64`, `riscv32`, `riscv64`, `x86-64`, `x86-32`.

**2. "Redundant assertion in unit test" — No change.**

The negative `assert!(!message.contains("Aarch64") && !message.contains("X86_64"))` is indeed logically subsumed by the exact-match `assert_eq!` above it. It is being kept deliberately: it documents the specific regression being pinned (no Rust enum variant names should leak into the user-facing message), which is the point of this PR. The redundancy is harmless and behavior-neutral, so it isn't worth a code change or a tracking issue.

_This triage was made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree, reply and we'll re-evaluate._

pmatos:
Thanks — the follow-up review on `2762c23` was triaged again by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5), who agreed on the outcome:

**New suggestion — replace `expect(...)` with `.map(|v| v.get_name()).unwrap_or("unknown arch")` — No action taken.**

1. The `.expect()` was added at this review's own prior request (the earlier round on this PR recommended exactly `self.to_possible_value().expect("no hidden variants").get_name()`). Swapping it back out would be churn against that same suggestion.
2. More substantively, `.unwrap_or("unknown arch")` trades a fail-fast panic on a programmer error for a silent, meaningless string inside a user-facing diagnostic — a silent-failure anti-pattern. All five `CliArch` variants are user-selectable and none is `#[value(skip)]`, so `to_possible_value()` cannot return `None` today. If a skipped variant were ever added (implausible for a CLI arch selector), a panic with a clear message is the correct fail-fast — far better than printing a misleading `unknown arch` in an error the user is meant to act on. The bot itself labels this "low-risk… suggestion rather than a blocker."

**Redundant negative assertions — No change (as noted previously).** Kept deliberately to document the exact regression being pinned (no Rust enum variant names leaking into the user-facing message); harmless and behavior-neutral.

_This triage was made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree, reply and we'll re-evaluate._

PR #477, opened 2026-06-18
fix(llm): count only real SMT queries

### LINKED ISSUE #212: LLM statistics count fast-path refutations as SMT queries
**Severity:** low  
**Category:** api-contract  
**Confidence:** medium  
**Triage:** contract-mismatch  
**Feature:** LLM Search Integration

## Description

SearchStatistics defines smt_queries as actual SMT queries, and LlmTimings already distinguishes real SMT calls via metrics.smt_called. The LLM loop nevertheless increments SearchStatistics.smt_queries for every equivalence failure or unknown outcome, including fast-path random refutations where no solver ran. This makes user-facing summaries and success rates overstate solver usage for LLM search. Confidence is downgraded because an included test currently asserts this mismatched accounting.

## Evidence

- `src/search/result.rs:68-71` (SearchStatistics)
- `src/search/llm/mod.rs:186-197` (LlmSearch::search)
- `src/search/llm/mod.rs:252-259` (LlmSearch::search)
- `src/search/llm/mod.rs:511-527` (fake_codex_equiv_fail_counts_verification_but_not_smt_fast_path)

## Recommendation

Increment SearchStatistics.smt_queries only when EquivalenceMetrics.smt_called is true. If LLM search needs to report verification attempts, add a separate statistic or use LlmTimings.verifications rather than overloading smt_queries.

## Repro

The existing fake_codex_equiv_fail_counts_verification_but_not_smt_fast_path test demonstrates the mismatch: timings.smt_calls is 0 while statistics().smt_queries is 1 for the same candidate.

The included test covers the behavior but treats the mismatch as expected. It should be updated to assert smt_queries == 0 for fast-path refutations and to check verifications separately.

## Suggested regression test

Change fake_codex_equiv_fail_counts_verification_but_not_smt_fast_path so a fast-path-only refutation leaves SearchStatistics.smt_queries at 0 while LlmTimings.verifications remains 1.

## Minimum fix scope

src/search/llm/mod.rs statistics updates and the affected LLM search tests.

---
Surfaced by clawpatch — finding id: `fnd_sig-feat-agent-tool-33938625af-a_368673d1c0`

pmatos:
Summary:
- Count LLM SearchStatistics::smt_queries only when equivalence metrics report a real SMT call.
- Keep fast-path random refutations counted as verifications without inflating SMT query totals.
- Add regression coverage for the fast-path and successful SMT-backed LLM paths.

Tests:
- cargo fmt --all
- cargo test search::llm::
- cargo test --all (after ./build_tests.sh; initial run failed only because binaries/ fixtures were absent)
- ./test_all.sh
- ./ci_check.sh

Closes #212

### DIFF
--- a/src/search/llm/mod.rs
+++ b/src/search/llm/mod.rs
@@ -210,6 +210,7 @@ impl SearchAlgorithm<crate::isa::AArch64> for LlmSearch {
                 timings.verify_time += verify_elapsed;
                 if m.smt_called {
                     timings.smt_calls += 1;
+                    stats.smt_queries += 1;
                     if let Some(bytes) = m.smt_formula_bytes {
                         timings.smt_formula_bytes_total += bytes;
                         if bytes > timings.smt_formula_bytes_max {
@@ -228,7 +229,6 @@ impl SearchAlgorithm<crate::isa::AArch64> for LlmSearch {
                             seq.len()
                         );
                     }
-                    stats.smt_queries += 1;
                     stats.smt_equivalent += 1;
                     stats.candidates_passed_fast += 1;
                     stats.improvements_found += 1;
@@ -272,13 +272,11 @@ impl SearchAlgorithm<crate::isa::AArch64> for LlmSearch {
                     }
                 }
                 IterationOutcome::EquivFail => {
-                    stats.smt_queries += 1;
                     if config.verbose {
                         eprintln!("llm-search: equiv-fail on call {}", call_idx);
                     }
                 }
                 IterationOutcome::EquivUnknown => {
-                    stats.smt_queries += 1;
                     if config.verbose {
                         eprintln!("llm-search: equiv-unknown on call {}", call_idx);
                     }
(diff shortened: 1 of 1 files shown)


pmatos:
**Deferred (diminishing returns)** — Regarding the automated code-review summary above (the two non-blocking "Minor Observations"):

The only concrete ask is adding a doc comment to `SearchStatistics::smt_queries` clarifying it counts actual Z3 invocations, not fast-path refutations. The ask is legitimate and even matches the file's own convention (the sibling `smt_elapsed` field already carries a precise multi-line doc comment), but it is a pure documentation polish that the review itself routed to a follow-up and marked non-blocking. This PR is fully green and ready to merge; making a one-line doc change now would re-trigger the full (~40 min) CI matrix for no functional benefit, and a clarifying comment on a cross-algorithm field is cleanly separable into its own small change. The second observation (the old `smt_queries` increment under `Success` being "subtly redundant") is a pure observation with no action requested.

Tracked as #504 (https://github.com/pmatos/s11/issues/504).

_Triaged by an independent AI reviewer (Claude Opus 4.8); the cross-harness second opinion could not run this round (Codex auth expired), so the conservative default of DEFER applied. If you disagree, reply here and it will be re-evaluated._

PR #478, opened 2026-06-18
fix(symbolic): enforce configured cost and window bounds

### LINKED ISSUE #222: Symbolic cost/window limits are accepted but never enforced
**Severity:** medium  
**Category:** api-contract  
**Confidence:** high  
**Triage:** contract-mismatch  
**Feature:** Symbolic Search

## Description

SymbolicConfig exposes a maximum synthesis window and an initial cost bound, but SymbolicSearch initializes best_cost from the original sequence and iterates candidate lengths up to target.len() without reading either field. A library caller can set an impossible cost bound, or a smaller window, and the search will still evaluate and return candidates outside that declared contract.

## Evidence

- `src/search/config.rs:212-222` (SymbolicConfig)
- `src/search/symbolic/synthesis.rs:47-52` (SymbolicSearch::linear_search)
- `src/search/symbolic/synthesis.rs:303-308` (SymbolicSearch::search)

## Recommendation

Apply config.symbolic.cost_bound when initializing best_cost, define whether it is strict or inclusive, and enforce or validate window_size in the search range. If these knobs are not supported, remove them from the public config instead of silently accepting them.

## Repro

Use the mov_add_sequence unit-test target with SearchConfig::default().with_symbolic(SymbolicConfig::default().with_cost_bound(0)); current code still enters length-1 search and can return the ADD rewrite because best_cost is initialized from original_cost instead of the configured bound.

The config tests only assert that the builder stores window_size and cost_bound; the SymbolicSearch behavior tests do not set either knob.

## Suggested regression test

Add a SymbolicSearch unit test that sets cost_bound to 0 for a known optimizable two-instruction target and asserts no optimization is returned, plus a window_size test once the intended window semantics are defined.

## Minimum fix scope

src/search/symbolic/synthesis.rs linear_search/search setup and, if semantics need clarification, src/search/config.rs docs/tests.

---
Surfaced by clawpatch — finding id: `fnd_sig-feat-library-0c23e80da9-a219_e089b928f3`

pmatos:
Summary:
- enforce SymbolicConfig::cost_bound as an exclusive pruning threshold capped by original target cost
- cap symbolic linear search lengths by SymbolicConfig::window_size
- document the config semantics and add regressions for cost/window enforcement

Tests:
- cargo fmt --all
- cargo test --all
- ./ci_check.sh
- scripts/full_test.sh: not present in this s11 workspace

Closes #222

### DIFF
--- a/src/search/config.rs
+++ b/src/search/config.rs
@@ -214,9 +214,16 @@ impl std::str::FromStr for SearchMode {
 /// Configuration for symbolic (SMT) search
 #[derive(Debug, Clone)]
 pub struct SymbolicConfig {
-    /// Maximum window size for synthesis
+    /// Maximum number of synthesized non-terminator instructions to consider.
+    ///
+    /// A value of 0 disables candidate search. If the target ends in a fixed
+    /// terminator, that terminator is appended after synthesis and does not
+    /// count against this window.
     pub window_size: usize,
-    /// Initial cost bound (None = use target cost)
+    /// Exclusive initial cost bound.
+    ///
+    /// Candidate sequences must be strictly cheaper than this bound and the
+    /// original target cost. `None` uses the original target cost.
     pub cost_bound: Option<u64>,
     /// Search mode (linear or binary)
     pub search_mode: SearchMode,
--- a/src/search/symbolic/synthesis.rs
+++ b/src/search/symbolic/synthesis.rs
@@ -2,8 +2,8 @@
 //!
 //! This module implements symbolic search using Z3 for equivalence verification.
 //! The approach uses linear cost search: try sequences of length 1, 2, ... up to
-//! the target length - 1, and for each length, enumerate candidates and verify
-//! equivalence with SMT.
+//! the configured synthesis window and target length - 1, and for each length,
+//! enumerate candidates and verify equivalence with SMT.
 //!
 //! Note: Full symbolic synthesis with symbolic opcodes/operands is very complex.
 //! This implementation uses a hybrid approach: enumerate concrete candidates
@@ -60,7 +60,8 @@ impl<I> SymbolicSearch<I>
 where
     I: ISA + SymbolicBackend<I>,
 {
-    /// Linear cost search: try each length from 1 to target length - 1
+    /// Linear cost search: try each length from 1 to the configured window,
+    /// capped at target length - 1.
     fn linear_search(
         &mut self,
         target: &[I::Instruction],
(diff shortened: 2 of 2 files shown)


pmatos:
**Triage of the automated review** — each item was assessed by two independent AI reviewers (Claude Opus 4.8 and Codex).

**Fixed in `9b75b27`** — Suggestion 2 (cost_bound=0 implicit invariant).
- Changed: added an inline comment in `symbolic_cost_bound_zero_prevents_known_mov_add_rewrite` (`src/search/symbolic/synthesis.rs`) documenting the load-bearing invariant — every AArch64 instruction cost is ≥ 1, so `best_cost = 0` prunes every candidate before any SMT query. The test is now self-contained.
- Validation: `./ci_check.sh` (fmt check + build + unit tests + test-binary builds + full suite) passed green.

**Deferred → #518** — three lower-value polish items, tracked together to avoid cosmetic churn on an already-green PR:
- *Behaviour-change flag* (`window_size = 3` now actively enforced): worth a note, but the field doc was already rewritten here and the enforcement is the PR's stated point; reviewers split (FIX/DEFER) → deferred.
- *Suggestion 1* (combined `cost_bound` + `window_size` test): the two code paths are orthogonal (`best_cost` init vs. `max_len` loop bound, no shared state) and each is already covered; both reviewers DEFER as future hardening.
- *Suggestion 4* (`Some(0)` doc note): the existing "strictly cheaper than this bound" wording already implies it; reviewers split (FIX/DEFER) → deferred.

**No change** — Suggestion 3 (rename `statistics.best_cost_found` → `original_cost`/`baseline_cost`): you scoped this explicitly as a follow-up on a pre-existing field that's set across the statistics struct, so it's out of scope for this bug-fix PR. Both reviewers REJECT.

_The DEFER/REJECT calls were made by two independent AI reviewers (Claude Opus 4.8 and Codex). If you disagree on any item, reply here and we'll re-evaluate._

pmatos:
**Triage of the re-review** ([#issuecomment-4746889076](https://github.com/pmatos/s11/pull/478#issuecomment-4746889076)) — each item assessed by two independent AI reviewers (Claude Opus 4.8 and Codex). This review is non-blocking ("Approve with minor suggestions"); none warranted a code change, so there is no further push.

**Deferred → #518** (folded in alongside the earlier follow-ups):
- *Loose-bound doc/log note* — `cost_bound > original_cost` is silently a no-op (`bound.min(original_cost)`). Correct, but a non-blocking doc/observability nit; tracked with the other deferred doc notes.
- *`without_cost_bound()` builder* — speculative API surface (gated on "if the builder pattern grows"); no current caller needs to reset a bound to `None`.
- *Loose-bound test* (`cost_bound > original_cost` treated as no-bound) — a legitimate in-scope test gap, batched with the combined-test deferral on #518.

**No change:**
- *Reword "initial" in the `cost_bound` doc* — the doc body already disambiguates ("strictly cheaper than this bound **and** the original target cost"); a one-word style preference with no project standard behind it.
- *Module-comment length / `best_cost_found` naming* — both self-labeled non-blocking; the comment documents the non-obvious hybrid enumerate-then-SMT approach, and `best_cost_found` is the same pre-existing, out-of-scope field already addressed in the previous triage.

_Outcomes decided by two independent AI reviewers (Claude Opus 4.8 and Codex). Reply here if you'd like any deferred item pulled into this PR instead._

PR #480, opened 2026-06-18
fix(x86): reject width-erasing register aliases

### LINKED ISSUE #226: x86 register aliases erase operand width and reassemble as full-width operations
**Severity:** high  
**Category:** bug  
**Confidence:** high  
**Triage:** confirmed-bug  
**Feature:** x86 Backend Semantics

## Description

The Capstone-to-IR path accepts `eax`, `ax`, and `al`, but the IR stores only `X86Register` with no operand-width field. Concrete and SMT semantics then use the ELF mode width, and the assembler emits `Rq` in x86-64 mode. This turns 32-bit zero-extending operations and 8/16-bit partial-register operations into 64-bit operations. For example, x86-64 `add eax, 0` zero-extends the high 32 bits and sets 32-bit flags; this implementation models and reassembles it as `add rax, 0`, preserving high 32 bits and using 64-bit flags.

## Evidence

- `src/main.rs:920-941` (parse_x86_register)
- `src/isa/x86.rs:160-188` (X86Instruction)
- `src/semantics/concrete_x86.rs:107-130` (apply_binop)
- `src/assembler/x86.rs:128-132` (encode_64)
- `src/main.rs:1991-2014` (x86_register_parser_covers_all_alias_groups)

## Recommendation

Until operand widths are represented end to end, reject non-mode-width aliases in the binary parser. The complete fix is to add operand width to the x86 IR and thread it through concrete semantics, SMT, cost, candidate generation, and assembly, including x86-64 32-bit zero-extension and 8/16-bit partial-write behavior.

## Repro

An x86-64 window containing two `add eax, 0` instructions is converted to two full-width `AddImm { rd: RAX, imm: 0 }` instructions and can be shortened to one full-width `add rax, 0`. With initial RAX=0x0000000100000000, the original 32-bit operations leave RAX=0, while the replacement leaves RAX=0x0000000100000000.

The parser tests explicitly lock in alias acceptance, but no test checks the architectural semantics or emitted bytes for `eax`/`ax`/`al` in x86-64 optimization. The linked opt integration test uses AArch64 fixture paths.

## Suggested regression test

Add x86-64 equivalence/optimization tests for `add eax, 0; add eax, 0` with high 32 bits live, and for `mov al, imm`, asserting these are rejected or modeled with the correct operand width.

## Minimum fix scope

`parse_x86_register`/`convert_to_x86_ir` plus x86 IR width representation; if rejecting aliases, the parser and tests are enough for the minimal safe fix.

---
Surfaced by clawpatch — finding id: `fnd_sig-feat-library-0e7c8afeb7-ce41_6d880ca4fc`

pmatos:
Summary:
- add a mode-aware x86 parser entry point that rejects non-mode-width register aliases from binary input
- thread x86-64/x86-32 mode into Capstone-to-IR conversion and cover raw add/mov alias regressions
- document the temporary mode-width-only x86 binary optimization restriction

Tests:
- cargo fmt --all
- cargo test --all (after ./build_tests.sh populated integration fixtures)
- ./ci_check.sh
- scripts/full_test.sh was requested by the generic run contract but does not exist in this repository; ./ci_check.sh ran the repo's test_all.sh instead

Closes #226

### DIFF
--- a/docs/capability.md
+++ b/docs/capability.md
@@ -89,6 +89,13 @@ The data-movement/arithmetic/logical/comparison families have register and
 immediate forms where the x86 IR models them. `cmov<cond>` has register
 operands and reads EFLAGS without modifying them.
 
+The x86 IR does not yet carry operand width. To avoid rewriting partial-width
+operations as full-width operations, the binary optimization path currently
+accepts only mode-width register aliases: `rax`/`r8`-style 64-bit names for
+x86-64, and `eax`-style 32-bit i386 names for x86-32. x86-64 `eax`/`ax`/`al`
+forms, x86-32 `ax`/`al` forms, and x86-32 extended-register aliases such as
+`r8d` are rejected until operand width is represented end to end.
+
 Fixed control-flow terminators:
 
 - `j<cond>` — parsed as an opaque trailing terminator and held fixed. Search
--- a/src/main.rs
+++ b/src/main.rs
@@ -563,6 +563,14 @@ impl X86OptimizationBackend {
         };
         Ok(Self { arch, width })
     }
+
+    fn parse_mode(&self) -> parser::x86::X86ParseMode {
+        match self.arch {
+            DetectedArch::X86_64 => parser::x86::X86ParseMode::Mode64,
+            DetectedArch::X86_32 => parser::x86::X86ParseMode::Mode32,
+            DetectedArch::Aarch64 => unreachable!("x86 backend never receives AArch64"),
+        }
+    }
 }
 
 impl ElfOptimizationBackend for X86OptimizationBackend {
@@ -596,7 +604,7 @@ impl ElfOptimizationBackend for X86OptimizationBackend {
         &self,
         instructions: &capstone::Instructions,
     ) -> Result<Vec<Self::Instruction>, String> {
-        convert_to_x86_ir(instructions)
+        convert_to_x86_ir(instructions, self.parse_mode())
     }
 
     fn validate_window_ir(&self, ir: &[Self::Instruction]) -> Result<(), String> {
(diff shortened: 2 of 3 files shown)


pmatos:
## Autofix triage

All CI is green and there are no inline review threads. The two automated comments above were each triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Outcomes:

### Automated code review — the four optional suggestions

1. **Rename/split the misleading test name** — **Deferred (diminishing returns).** Zero-behavior cosmetic rename on a green PR; tracked in #516 rather than churned here.
2. **Comment documenting the `ah`/`bh`/`ch`/`dh` omission** — **Already handled (no change).** The intentional omission is already documented at `src/parser/x86.rs:157-160` on `parse_x86_register`: *"Legacy high-byte aliases (`ah`, `bh`, `ch`, `dh`) are intentionally excluded because the minimal x86 IR models the low-byte/REX alias set."* `classify_x86_register_alias` is the helper that function delegates to, so the existing doc covers it; duplicating the note would also conflict with the repo's "comment sparingly" convention.
3. **Collapse the `_with_mode` / `_for_mode` wrappers** — **Deferred (out of scope).** Reviewer flagged it as "Not a blocker"; it's a pure refactor of working, tested code. Tracked in #516.
4. **Comment on the guard ordering (`r8d` in Mode32)** — **Deferred (diminishing returns).** The branch order is correct and already pinned by a test. Tracked in #516.

Deferred items 1, 3, and 4 are tracked together in #516.

### Codecov report

**No action taken (informational).** Patch coverage is 98.97% with both `codecov/patch` and `codecov/project` green, and project coverage rose +0.05%. The new parser behavior is directly exercised by the added tests, so the 3 uncovered lines are not actionable.

_This assessment was made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree with any outcome, reply and we'll re-evaluate._

PR #488, opened 2026-06-18
fix(search): count SMT-reaching candidates as fast passes

### LINKED ISSUE #238: Verification statistics count non-SMT paths as SMT work
**Severity:** low  
**Category:** bug  
**Confidence:** high  
**Triage:** confirmed-bug  
**Feature:** Enumerative Search

## Description

`verify_candidate` increments `smt_queries` before calling the equivalence checker and only subtracts for `NotEquivalentFast`. But `check_equivalence_with_config` can return `NotEquivalent` from `pre_smt_guard` before fast-path or SMT, so those candidates are reported as SMT queries even though no solver ran. Conversely, candidates that pass fast validation but are disproved by SMT also return `NotEquivalent`, so `candidates_passed_fast` is not updated for them.

## Evidence

- `src/search/result.rs:64-71` (SearchStatistics)
- `src/search/enumerative/search.rs:94-105` (verify_candidate)
- `src/semantics/equivalence.rs:378-387` (check_equivalence_with_config)

## Recommendation

Use `check_equivalence_with_config_metrics` or extend the equivalence API so enumerative search updates `smt_queries` from an actual `smt_called` flag and increments `candidates_passed_fast` for candidates that reach SMT, regardless of the SMT result.

## Repro

A target that writes NZCV compared against a candidate that does not write NZCV returns `NotEquivalent` from `pre_smt_guard` with `flags_live=true`; `verify_candidate` still leaves `smt_queries` incremented.

The enumerative tests assert optimization outcomes and timeout shape, but do not assert statistics for pre-SMT rejection or fast-pass-then-SMT-fail cases.

## Suggested regression test

Add a module test that calls `verify_candidate` on a pre-SMT flag-writer divergence and asserts `smt_queries == 0`; add another where fast path passes but SMT returns `NotEquivalent` and assert `candidates_passed_fast` increments.

## Minimum fix scope

Change `verify_candidate` accounting in `src/search/enumerative/search.rs` and add focused statistics tests.

---
Surfaced by clawpatch — finding id: `fnd_sig-feat-library-ade3027169-d1ab_43eedaa2b0`

pmatos:
Summary:
- Count enumerative `candidates_passed_fast` from `metrics.smt_called` so SMT refutations are included without counting pre-SMT rejections.
- Add regressions for pre-SMT flag-writer rejection and fast-pass-then-SMT-fail statistics.

Verification:
- cargo fmt --all
- ./build_tests.sh
- cargo test --all (after building test binaries; the first run failed only because binaries/ fixtures were missing)
- ./ci_check.sh
- `scripts/full_test.sh` was requested by the generic run contract but does not exist in this s11 repository. The repo-specific `./ci_check.sh` ran `./test_all.sh`.

Closes #238

### DIFF
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -271,7 +271,11 @@ where
         .fetch_add(solver_nanos, Ordering::Relaxed);
     // Count only candidates that actually reached `solver.check()`. The same
     // metric also marks candidates that passed fast concrete validation,
-    // including ones later disproved by SMT.
+    // including ones later disproved by SMT. `smt_called` is the right guard
+    // because every early return that never reaches the solver leaves it
+    // false: the pre-SMT guard (`flag_writers_diverge && flags_live`) returns
+    // `NotEquivalent` and the fast-path returns `NotEquivalentFast`, neither of
+    // which should count as a fast pass (PR #269 review).
     if metrics.smt_called {
         shared.smt_queries.fetch_add(1, Ordering::Relaxed);
         shared
@@ -1354,6 +1358,39 @@ mod tests {
         assert_eq!(shared.smt_queries.load(Ordering::Relaxed), 1);
         assert_eq!(shared.candidates_passed_fast.load(Ordering::Relaxed), 1);
         assert_eq!(shared.smt_equivalent.load(Ordering::Relaxed), 0);
+    }
+
+    #[test]
+    fn verify_candidate_counts_smt_equivalence() {
+        // Commutativity: `add x0, x1, x2` == `add x0, x2, x1`. The candidate
+        // passes fast concrete validation and Z3 then proves equivalence, so
+        // the success path must record one SMT query, one fast pass, and one
+        // equivalence.
+        let target = vec![Instruction::Add {
+            rd: Register::X0,
+            rn: Register::X1,
+            rm: Operand::Register(Register::X2),
+        }];
+        let candidate = vec![Instruction::Add {
+            rd: Register::X0,
+            rn: Register::X2,
+            rm: Operand::Register(Register::X1),
+        }];
+        let live_out = LiveOut::from_registers(vec![Register::X0]);
+        let config = SearchConfig::default().with_timeout_option(None);
+        let shared = SharedState::<AArch64>::new(u64::MAX);
+
+        assert!(verify_candidate::<AArch64>(
+            &target,
+            &candidate,
+            &live_out,
+            &config,
+            &shared,
+            std::time::Instant::now(),
+        ));
+        assert_eq!(shared.smt_queries.load(Ordering::Relaxed), 1);
+        assert_eq!(shared.candidates_passed_fast.load(Ordering::Relaxed), 1);
+        assert_eq!(shared.smt_equivalent.load(Ordering::Relaxed), 1);
     }
 
     #[test]
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -269,21 +269,18 @@ where
     shared
         .smt_elapsed_nanos
         .fetch_add(solver_nanos, Ordering::Relaxed);
-    // Count only candidates that actually reached `solver.check()`. The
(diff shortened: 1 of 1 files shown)


pmatos:
Autonomous run note:
- The generic run contract mentioned `scripts/full_test.sh`, but this s11 checkout has no such script. I ran the repo gates instead: `cargo fmt --all`, `cargo clippy --all -- -D warnings`, `./build_tests.sh`, `cargo test --all`, and `./ci_check.sh` (which also runs `./test_all.sh`).
- The issue carried `ready-for-agent`, not `agent-ready`, so I removed `ready-for-agent` as the readiness label after opening this PR.

pmatos:
Thanks for the review. Each suggestion was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5); they agreed on every item.

**1. Missing success-path test — fixed in `e3a411a`.**
Added `verify_candidate_counts_smt_equivalence` using a commutative pair (`add x0, x1, x2` ≡ `add x0, x2, x1`), asserting the success path records 1 SMT query, 1 fast pass, and 1 equivalence. The enumerative statistics tests are now symmetric across the pre-SMT-guard, refutation, and equivalence cases.
Validation: `cargo test --lib verify_candidate` plus the full `./ci_check.sh` gate (fmt, build, 1127 unit/integration tests, `test_all.sh`) — all green.

**2. Comment that the two counters move together — no change (style preference).**
The two `fetch_add` calls sit adjacent inside the single `if metrics.smt_called` block, so their local coupling is already self-evident; the cross-path divergence lives in the stochastic/symbolic/llm increments rather than this function. Per the project guideline to comment sparingly, an extra note here is diminishing-returns churn.

**3. Restored the "why" in the comment — fixed in `e3a411a`.**
The comment now explains that `smt_called` is the right guard precisely because every early return that never reaches the solver — the pre-SMT guard (`flag_writers_diverge && flags_live`) and the fast-path — leaves it `false`, so those candidates are correctly excluded from the fast-pass count.

**4. `fast_pass_rate()` semantic change — no action.**
Agreed this is informational; the shift to a true fast-validation pass rate is the intended fix for #238, and as noted no PR action is required.

_Items 2 and 4 were assessed by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree with either, reply and we'll re-evaluate._

pmatos:
Thanks for the follow-up review — triaged again by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5), who agreed on all five.

**1. Rename `verify_candidate_counts_smt_refutation_as_fast_pass` — no change (style preference).**
The current name reads as "counts a fast-passing, SMT-refuted candidate," which is what it asserts; renaming is a naming preference with no project backing.

**2. Add `smt_elapsed_nanos > 0` to the refutation test — no change.**
`smt_queries == 1` already proves the solver ran; a positive wall-clock assertion adds flake risk (a very fast `check()` can round to 0 ns) for marginal value.

**3. Re-word the comment — no change.**
This is the same comment the previous review asked us to expand with the pre-SMT-guard rationale (done in `e3a411a`). It's accurate, and per the project guideline to comment sparingly, re-wording it again is diminishing-returns churn.

**4. Cross-algorithm inconsistency in `synthesis.rs` / `llm/mod.rs` — deferred, tracked as #502.**
Verified: the symbolic and LLM paths increment `candidates_passed_fast` only on the `Equivalent` arm, so fast-passing SMT-refuted candidates are undercounted — the same class of bug as #238, but in code outside this PR's enumerative scope. Filed as #502.

**5. `smt_queries` underflow in `synthesis.rs` — no change (not an issue).**
`smt_queries` is incremented unconditionally at the top of the equivalence call, before the `NotEquivalentFast` arm's `-= 1` can run, so the counter is always ≥ 1 there and cannot underflow. The increment-then-decrement pattern is intentional (don't count fast-rejected candidates as SMT queries).

No code change this round; all items are style/no-op except #4, tracked separately as #502.

_Items 1–3 and 5 were assessed by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree with any, reply and we'll re-evaluate._

PR #489, opened 2026-06-18
fix(search): enumerate longer candidate products

### LINKED ISSUE #239: Enumerative search silently skips all lengths above two
**Severity:** medium  
**Category:** bug  
**Confidence:** high  
**Triage:** confirmed-bug  
**Feature:** Enumerative Search

## Description

The module-level contract says it enumerates every shorter length, but the dispatcher only runs length 1 and 2 and treats length >= 3 as a no-op. A length-4 target with three independent live register writes can have a valid length-3 rewrite and no possible length-1/2 rewrite; this implementation will return no optimization despite the candidate existing in the pool.

## Evidence

- `src/search/enumerative/search.rs:3-6`
- `src/search/enumerative/search.rs:229-236` (EnumerativeSearch::search)
- `src/search/enumerative/search.rs:496-536` (tests::length_four_target_iterates_past_length_three_dispatch)

## Recommendation

Implement generic/product enumeration for every length in `1..target.len()` with pruning and timeout checks, or narrow the public/documented contract to length <= 2 if that is the intended feature boundary.

## Repro

Use a length-4 target that zeroes X0, X2, and X3, with live_out={X0,X2,X3}; the length-3 sequence `mov x0,#0; mov x2,#0; mov x3,#0` is strictly shorter, but search never dispatches length 3.

The length-4 unit test intentionally uses a target with a length-1 collapse, so it proves only that the length-3 no-op arm does not panic, not that length-3 candidates are searched.

## Suggested regression test

Add an enumerative unit test with a length-4 target whose only shorter equivalent in the configured pool has length 3, and assert `optimized_sequence.len() == 3`.

## Minimum fix scope

Update `src/search/enumerative/search.rs` length dispatch/enumeration and add the targeted regression test in the same module.

---
Surfaced by clawpatch — finding id: `fnd_sig-feat-library-ade3027169-db63_c6238fb5a8`

pmatos:
Summary:
- Add generic product enumeration for candidate lengths 3..target.len()-1.
- Share candidate finalization so terminator handling, cost pruning, stats, and equivalence checks stay consistent across lengths.
- Add a length lower-bound skip to avoid exhaustive product sweeps once the current best cannot be beaten.
- Add regressions for a length-4 target whose only rewrite is length 3 and for nested product timeout handling.

Tests:
- cargo fmt --all -- --check
- cargo test search::enumerative
- cargo test --all (after ./build_tests.sh; first run failed only because fixture binaries were absent)
- ./ci_check.sh

Note: scripts/full_test.sh was requested by the autonomous contract but does not exist in this repo.

Closes #239

### DIFF
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -68,7 +68,7 @@ impl<I: ISA> SharedState<I> {
         // sequentially (worse first, better second) and then have the worse
         // thread acquire the mutex last and clobber the better candidate.
         //
-        // Ordering: the cost-prune fast path in `run_length_one`/`_two` loads
+        // Ordering: the cost-prune fast path in the length runners loads
         // `best_cost` with `Ordering::Acquire`. The store below is `Release`
         // (not `Relaxed`) so those loads observe the new value — the mutex
         // unlock would publish the mutex-protected `best`, but the atomic is
@@ -424,6 +424,49 @@ fn candidate_solver_timeout(config: &SearchConfig, start: Instant) -> Option<Dur
     candidate_solver_timeout_for_elapsed(config, start.elapsed())
 }
 
+fn evaluate_candidate<I>(
+    target: &[I::Instruction],
+    live_out: &<I as EnumerativeBackend<I>>::LiveOut,
+    config: &SearchConfig,
+    mut candidate: Vec<I::Instruction>,
+    terminator: Option<I::Instruction>,
+    shared: &SharedState<I>,
+    start: Instant,
+) where
+    I: ISA + EnumerativeBackend<I>,
+{
+    if let Some(t) = terminator {
+        candidate.push(t);
+    }
+    let candidate_cost = <I as EnumerativeBackend<I>>::sequence_cost(&candidate, config);
+    if candidate_cost >= shared.best_cost.load(Ordering::Acquire) {
+        return;
+    }
+    shared.candidates_evaluated.fetch_add(1, Ordering::Relaxed);
+    if verify_candidate::<I>(target, &candidate, live_out, config, shared, start) {
+        shared.record_improvement(candidate, candidate_cost);
+    }
+}
+
+fn minimum_generated_instruction_cost<I>(
+    config: &SearchConfig,
+    all_instructions: &[I::Instruction],
+) -> Option<u64>
+where
+    I: ISA + EnumerativeBackend<I>,
+{
+    all_instructions
+        .iter()
+        .map(|instr| <I as EnumerativeBackend<I>>::sequence_cost(&[*instr], config))
+        .min()
+}
+
+fn length_cost_lower_bound(length: usize, min_instruction_cost: u64, terminator_cost: u64) -> u64 {
+    min_instruction_cost
+        .saturating_mul(length as u64)
+        .saturating_add(terminator_cost)
+}
+
 fn run_length_one<I>(
     target: &[I::Instruction],
     live_out: &<I as EnumerativeBackend<I>>::LiveOut,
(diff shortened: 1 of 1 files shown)


pmatos:
Implementation note: I used generic product enumeration for lengths >= 3, plus a conservative length-level cost lower bound before dispatch. The alternative was to rely only on per-candidate cost pruning, but the existing length-4/length-1-collapse test exposed that this can still enumerate a huge length-3 product with no timeout. The lower bound preserves the documented “strictly cheaper” search behavior while skipping lengths that cannot beat the current best.

pmatos:
Regarding the automated code review above — each item was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Outcomes:

**1. `continue` → `break` in the lower-bound skip — Fixed in `31d9121`.**
Changed: `run_lengths` now `break`s once `lower_bound >= best_cost`. The per-length lower bound is non-decreasing in `length` and `best_cost` only falls, so once one length cannot beat the best, no longer length can either; the comment now states the monotonicity invariant. (Minor note: `continue` was not actually a correctness bug — it produced identical results, just scanned the remaining lengths paying an atomic load each.)
Validation: `./ci_check.sh` (fmt + clippy + build + full test suite) — all green.

**2. Redundant `stop_if_timed_out()` at the `enumerate_suffix` leaf — Fixed in `31d9121`.**
Changed: removed check (B). On the `candidate.len() == self.length` path it ran immediately after the entry check (A) with no intervening work, so it could never fire independently.
Validation: same `./ci_check.sh` run.

**3. Doc comment on `length_cost_lower_bound`'s additivity precondition — Deferred (diminishing returns).**
The additivity precondition holds for every current cost model (AArch64 instruction count, x86 CodeSize), so the doc only guards a hypothetical future non-additive model (e.g. macro-fusion). Legitimate but not load-bearing in this PR, and the two reviewers split FIX/DEFER on it. Tracked as #517 (https://github.com/pmatos/s11/issues/517).

**4. Replace the `context.length == 0` guard with `debug_assert!` — No change.**
Style preference with no clear win: the guard is dead but harmless and returns gracefully in release, whereas `debug_assert!` is a no-op in release. Both reviewers rejected swapping safe-return semantics for a debug-only assertion.

_This triage was made by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). If you disagree on any item, reply and we'll re-evaluate._

pmatos:
Second automated review triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Outcomes:

**Super-additivity doc on `sequence_cost` / `length_cost_lower_bound` — Already tracked.**
Same additivity-precondition concern as the previous review; already deferred and tracked as #517. No new action.

**`length == 0` guard → `debug_assert!` — No change (as before).**
Already addressed last round: the guard returns gracefully in release whereas `debug_assert!` is a no-op there, so the swap trades safe behaviour for a debug-only check. Both reviewers rejected it again.

**Tighten the `cost_calls < 8` timeout-test assertion — Deferred (diminishing returns).**
The trace is deterministic (`cost_calls == 1`, same 2× timing margin as the `== 1` sibling test), so tightening is a legitimate consistency win — but it is a test-only assertion the *first* review of this PR explicitly called "not a new problem," and it is not worth re-triggering a full CI/review round on an already-green PR. Tracked as #530 (https://github.com/pmatos/s11/issues/530).

**Coverage test for the empty-instruction-set (`min_instruction_cost == None`) branch — Deferred (diminishing returns).**
Legitimate but a marginal coverage nit on a trivial guard; `codecov/patch` is green (informational). Tracked as #531 (https://github.com/pmatos/s11/issues/531).

_Triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Reply if you disagree on any item and we'll re-evaluate._

PR #491, opened 2026-06-18
fix(parser): restrict memory extend shifts

### LINKED ISSUE #302: Restrict parse_memory_extend_tail to encodable extend shift set (sibling of #295)
Deferred from #289 (chatgpt-codex-connector[bot] thread [r3261695336](https://github.com/pmatos/s11/pull/289#discussion_r3261695336)): parser over-permissive on memory-op extend shift.

Original feedback by @chatgpt-codex-connector on PR #289:

> `parse_memory_extend_tail` accepts extend shifts in `0..=4` unconditionally, but AArch64 memory register-extend addressing only allows no shift or the access-size shift (and which value is legal depends on the load/store width). As written, operands like `ldrb ..., [.., uxtw #4]` pass parsing and `is_encodable_aarch64`, creating IR that does not correspond to a real encodable instruction and can fail later at emission.

**Context:** `src/parser/mod.rs:915` — `parse_memory_extend_tail` is the extend-mode equivalent of `parse_lsl_amount` (#295). Like #295, the parser currently accepts the wider 0..=4 range and downstream `is_encodable_aarch64` would gate it — but `is_encodable_ldr_like` does not yet have a per-width extend-shift check, so the gate is partial.

**Why deferred:** `diminishing-returns` — current behaviour is sound today for the `Ext` paths that the assembler accepts (UXTW/SXTW shifts other than 0 or log2(width) hit the assembler error path). The change is parse-time UX plus tightening `is_encodable_ldr_like` to symmetrise with the parser. Bundling with #295 (LSL set) would close the whole class.

pmatos:
Summary:
- Restrict AArch64 memory extend shifts to 0 or the access-size scale at parse time.
- Mirror the same rule in is_encodable_aarch64 for hand-built IR.

Closes #302

Verification:
- cargo fmt --all
- cargo test --all
- ./ci_check.sh
- Note: scripts/full_test.sh is not present in this s11 repo; ./ci_check.sh builds test binaries and runs ./test_all.sh.

### DIFF
--- a/src/ir/instructions.rs
+++ b/src/ir/instructions.rs
@@ -2,8 +2,8 @@
 
 use crate::ir::aarch64_encoding::logical_imm64_encodable;
 use crate::ir::types::{
-    AccessWidth, AddressOperand, Condition, IndexMode, LabelId, Operand, Register, RegisterWidth,
-    ShiftKind,
+    AccessWidth, AddressOperand, Condition, ExtendKind, IndexMode, LabelId, Operand, Register,
+    RegisterWidth, ShiftKind,
 };
 use std::fmt;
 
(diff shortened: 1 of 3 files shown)


pmatos:
@claude[bot] Regarding your code review (the two non-blocking nits):

**Deferred (diminishing returns)** — _negative-shift error message (Item 1)._ The rejection behavior is already correct (`uxtw #-1` is rejected); this is purely an error-message phrasing refinement. Two independent AI reviewers agreed it's a diminishing-returns nit rather than something to churn this PR for — ideally the clearer "out-of-range / negative" wording should be unified across both the LSL and extend paths. Tracked as #497 (https://github.com/pmatos/s11/issues/497).

**Style preference (no change)** — _`shift as u8` cast invariant (Item 2)._ The cast is reached only after the guard `shift == 0 || shift == i64::from(scaled_shift)`, where `scaled_shift ∈ {0,1,2,3}`, so it is provably in `u8` range two lines above. Both reviewers agreed a `debug_assert!`/comment would mostly restate the adjacent control flow and runs against this project's "comment sparingly; only comment complex code" rule — so no change.

PR #492, opened 2026-06-18
fix(equiv): warn when memory disables fast-only

### LINKED ISSUE #303: Surface a one-line warning when --fast-only is force-overridden by memory_live carve-out
Deferred from #289 (claude[bot] second review item 2): the ADR-0007 decision 5 carve-out silently clears `fast_only` for memory-bearing windows, but the ADR text says it should *also* log a one-line warning.

Original feedback by @claude[bot]:

> ADR-0007 decision 5 says: "force fast_only = false **and log a one-line warning**". The code in `check_equivalence_with_config` and `check_equivalence_with_config_metrics` silently clears `fast_only` with no user-visible warning. A user who explicitly passes `--fast-only` will have that flag silently ignored on any memory-bearing window. Even a single `eprintln!("[s11] warning: --fast-only disabled for memory-bearing window (see ADR-0007)")` would help.

**Context:** `src/semantics/equivalence.rs` — both `check_equivalence_with_config` (now a thin wrapper) and `check_equivalence_with_config_metrics` patch `config.fast_only = false` when `touches_memory(prefix1) || touches_memory(prefix2)` and the user had set it on.

**Why deferred:** `diminishing-returns` — correctness is intact (the carve-out is sound); the change is observability so users notice their flag was overridden. Two evaluators agree this is a UX improvement, not a correctness fix.

pmatos:
Summary:
- Emit the ADR-0007 warning when AArch64 memory-bearing equivalence checks force --fast-only off.
- Add CLI integration coverage for the memory warning and a register-only no-warning guard.

Closes #303

### DIFF
--- a/src/semantics/equivalence.rs
+++ b/src/semantics/equivalence.rs
@@ -378,9 +378,16 @@ impl EquivalenceBackend for AArch64 {
             || crate::validation::live_out::touches_memory(seq2);
         if memory_touched {
             if config.fast_only {
-                eprintln!(
-                    "[s11] warning: --fast-only disabled for memory-bearing window (see ADR-0007)"
-                );
+                // This carve-out fires once per equivalence check, and a search
+                // pass invokes the check per candidate/window. Emit the warning
+                // at most once per process so multi-window `opt --fast-only` runs
+                // don't flood stderr with identical lines (ADR-0007).
+                static FAST_ONLY_WARNED: std::sync::Once = std::sync::Once::new();
+                FAST_ONLY_WARNED.call_once(|| {
+                    eprintln!(
+                        "[s11] warning: --fast-only disabled for memory-bearing window (see ADR-0007)"
+                    );
+                });
             }
             config.memory_live = true;
             config.fast_only = false;
--- a/src/semantics/equivalence.rs
+++ b/src/semantics/equivalence.rs
@@ -377,6 +377,11 @@ impl EquivalenceBackend for AArch64 {
         let memory_touched = crate::validation::live_out::touches_memory(seq1)
             || crate::validation::live_out::touches_memory(seq2);
         if memory_touched {
+            if config.fast_only {
+                eprintln!(
+                    "[s11] warning: --fast-only disabled for memory-bearing window (see ADR-0007)"
+                );
+            }
             config.memory_live = true;
             config.fast_only = false;
         }
--- a/tests/integration/equiv_test.rs
+++ b/tests/integration/equiv_test.rs
@@ -1,0 +1,79 @@
+use std::process::{Command, Output};
+
+const FAST_ONLY_MEMORY_WARNING: &str =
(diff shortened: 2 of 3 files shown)


pmatos:
Regarding the automated code review on this PR — triaged with two independent evaluators (Claude Opus 4.8 and Codex gpt-5.5):

**1 & 2 — Warning flooding across memory-bearing windows → fixed in `0626eb9`.**
Changed: wrapped the ADR-0007 carve-out warning in `src/semantics/equivalence.rs` in a `std::sync::Once`, so it is emitted at most once per process. `check_equivalence_for_metrics` clones the config per call (`equivalence.rs:761`) and a search pass invokes the check per candidate/window, so a multi-window `opt --fast-only` run over memory-bearing code would otherwise repeat the line N times — this now matches #303's "one-line warning" intent.
Validation: `./ci_check.sh` green (fmt-check, build, unit + integration suites). The existing `equiv_fast_only_memory_window_warns_when_overridden` test still asserts exactly one emission per `s11 equiv` invocation.

**3 — No test for `--fast-only` absent + a memory instruction → deferred (diminishing returns).**
Real but strictly-additive regression coverage that the review marked low-priority; the gate is literally `if config.fast_only`, so the existing register-window test already proves the warning is conditional. Tracked as #507.

**Code-quality note — confirm `tempfile` in `[dev-dependencies]` → already handled, no change.**
`tempfile = "3.27"` is already under `[dev-dependencies]` in `Cargo.toml`, and the green `test` job proves the integration tests compile and run.

pmatos:
Follow-up automated review on `0626eb9`, triaged with two independent evaluators (Claude Opus 4.8 and Codex gpt-5.5) — consensus on all three findings:

**Finding 1 (comment block "violates CLAUDE.md one-short-line-max rule") → no change.**
No such rule exists in this repo — the only comment guidance is "use comments sparingly; only comment complex code" (no line limit). The 4-line comment documents a genuinely non-obvious invariant (why `Once` is needed given per-candidate invocation frequency), which is exactly the complexity that warrants a comment. Style preference with no project backing.

**Finding 2 (warning string duplicated prod↔test; share via `pub(crate)` const) → no change.**
The string lives in a separate integration-test crate, so "sharing" requires exporting a `pub` production constant across the crate boundary — which this review itself calls "slightly awkward," and which the *earlier* review praised as a "solid pattern." The test intentionally pins the user-visible stderr text; the duplication is the right trade-off here.

**Finding 3 (align `opt_test.rs` temp-dir with `tempfile`) → deferred.**
`opt_test.rs` is pre-existing and untouched by this PR; the review itself labels it "trivial / follow-up." Tracked as #512.

(Findings 4 & 5 were "correct" / "looks good" — no action.)

No code change this round — CI is green and the remaining items are a non-existent style rule, a contradictory cosmetic nit, and an out-of-scope cleanup, so there's nothing to churn.

PR #494, opened 2026-06-18
fix(parser): reject W-form signed-load destinations

### LINKED ISSUE #306: LDRSB/LDRSH/LDRSW W-form destination is silently treated as X-form (correctness)
Deferred from #289 (chatgpt-codex-connector[bot] P1 thread [r3266096162](https://github.com/pmatos/s11/pull/289#discussion_r3266096162)): the parser collapses W- and X-form destinations for signed loads into the same `Ldrs` IR, but downstream semantics always 64-bit-sign-extend.

Original feedback by @chatgpt-codex-connector on PR #289:

> When parsing valid `ldrsb wN, ...` or `ldrsh wN, ...` from Capstone or user input, this arm collapses the W/X destination spelling into the same `Ldrs` IR used for X-form loads. Downstream semantics and assembly then treat the instruction as the 64-bit sign-extending form (`ldrsb xN`/`ldrsh xN`), so a byte like `0xff` becomes `0xffffffffffffffff` instead of the architectural W-form result `0x00000000ffffffff`, and optimization can rewrite binaries with different register contents.

**Verified — the bug is real but currently dormant:**
- `parse_w_or_x_register` (src/parser/mod.rs:160) maps both `"w0"` and `"x0"` to `Register::X0`.
- `parse_single_reg_mem` (src/parser/mod.rs:940) only carries an `AccessWidth` — there is no destination-width bit on `Instruction::Ldrs`.
- `sign_extend_load` in `src/semantics/concrete.rs:667` always extends to 64 bits regardless of how the destination was spelled.
- The same applies in SMT (`src/semantics/smt.rs:988`, via `ldr_sign_extend(..., target_width=64)`).

For a byte `0xff`:
- `ldrsb xN, [...]` → `x[63:0] = 0xFFFF_FFFF_FFFF_FFFF` ✓ what we currently produce
- `ldrsb wN, [...]` → `w[31:0] = 0xFFFF_FFFF`, then writing W zeroes upper 32 → `x[63:0] = 0x0000_0000_FFFF_FFFF` ✗ we still emit the X-form result

Same dichotomy for `ldrsh wN`. `ldrsw` only has an X-form architecturally, but our parser would happily accept `ldrsw w0` and produce the same X-form result.

**Why dormant today:** no current test or tested ELF contains `ldrsb wN` / `ldrsh wN` (`grep -rn "ldrsb w\|ldrsh w\|ldrsw w"` in `src/` and `tests/` returns nothing). The Capstone-tripwire rows in #289 all use X-form destinations. So the silent miscompilation cannot trigger on today's inputs — but it is a sharp edge waiting on any future binary that disassembles to a W-form LDRS.

**Two correct fixes (pick one in the follow-up):**
1. **Surgical (parser guard).** Reject `ldrsb wN` / `ldrsh wN` / `ldrsw w*` at the parser; treat them as `Unsupported` for now. Low-risk; matches how the parser handles other W-destination forms the IR can't represent.
2. **Proper (IR change).** Add a destination-width bit (or split into `Ldrs32` / `Ldrs64`) and thread it through concrete + SMT + assembler. Higher-risk; touches semantics + tests.

**Why deferred:** the PR is large and the bug is dormant. Both fixes warrant their own review cycle. Filing high-priority so the next iteration picks this up before any future binary surfaces it.

pmatos:
Summary:
- Reject W-form destinations for AArch64 signed-load mnemonics before the parser erases W/X spelling into the shared register IR.
- Keep X-form `ldrsb` / `ldrsh` / `ldrsw` accepted across text and Capstone paths.
- Document the current X-form-only signed-load limitation in the capability matrix.

Tests:
- `cargo fmt --all`
- `cargo test --all`
- `./ci_check.sh`

Closes #306

### DIFF
--- a/docs/capability.md
+++ b/docs/capability.md
@@ -61,8 +61,12 @@ Rewritable straight-line mnemonics accepted by the parser and Capstone bridge:
   byte-addressed Z3-array memory model with sound full aliasing, whole-memory
   live-out auto-derived): `ldr`, `ldrb`, `ldrh`, `ldrsb`, `ldrsh`, `ldrsw`,
   `str`, `strb`, `strh`, `ldp`, `stp`, `ldpsw` — accepted in immediate-offset,
-  pre-index, post-index, register-offset, and register-extend addressing
-  forms, in both `W` and `X` widths
+  pre-index, post-index, register-offset, and register-extend addressing forms.
+  Unsized `ldr` / `str` infer `W` vs `X` width from the data register spelling.
+  Zero-extending `ldrb` / `ldrh` loads and `strb` / `strh` stores use scoped
+  `W`/`X` register slots.
+  `ldrsb` / `ldrsh` / `ldrsw` signed loads currently accept only X-form
+  destinations because the current `Ldrs` IR models X-form sign-extension.
 
 Fixed control-flow terminators:
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -2845,6 +2845,23 @@ mod cli_helper_tests {
         }
     }
 
+    #[test]
+    fn convert_capstone_op_rejects_w_form_signed_load_destinations() {
+        for (mnem, ops) in [
+            ("ldrsb", "w0, [x1]"),
+            ("ldrsh", "w0, [x1]"),
+            ("ldrsw", "w0, [x1]"),
+        ] {
+            match convert_capstone_op(mnem, ops) {
+                ConvertOutcome::Unsupported(line) => {
+                    assert!(line.contains(mnem));
+                    assert!(line.contains("X-form"));
+                }
+                other => panic!("expected Unsupported for `{mnem} {ops}`, got {other:?}"),
+            }
+        }
+    }
+
     fn assemble_aarch64_test_bytes(instructions: &[Instruction]) -> Vec<u8> {
         AArch64Assembler::new()
             .assemble_instructions(instructions, 0x1000)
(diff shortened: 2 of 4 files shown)


pmatos:
Regarding the automated `claude-review` summary (the three minor points; the review itself concluded "Ready to merge"):

**1. `operands.len() >= 2` vs `!operands.is_empty()` guard** — **Style preference (no change)** — The review itself notes "the current behaviour is correct." A single-operand `ldrsb w0` is still rejected by `parse_single_reg_mem` (operand-count check); the suggestion only changes which error message precedes the other. No functional difference, so no change on this green PR.

**2. Architectural note (`ldrsb w0`/`ldrsh w0` are valid AArch64)** — **Already handled (no change)** — Informational, and the docs already attribute the restriction to the IR, not the architecture: `docs/capability.md` says signed loads "currently accept only X-form destinations because the current `Ldrs` IR models X-form sign-extension." That is exactly the "current IR limitation, not an arch restriction" framing requested. The review also explicitly says "No changes needed here."

**3. Add an `ldrsw` addressing-mode case for test symmetry** — **Deferred (diminishing returns)** — Legitimate but low-value: the W/X destination guard in `parse_signed_load_mem` runs before any address parsing, so additional `ldrsw` addressing-mode cases exercise no distinct code path. Tracked for symmetry/readability rather than churning this focused fix. Tracked as #506 (https://github.com/pmatos/s11/issues/506).

PR #496, opened 2026-06-18
fix(parser): reject pair register-index addressing

### LINKED ISSUE #340: Reject register-offset addressing for LDP/STP pair instructions
Deferred from #333: LDP/STP silently accept architecturally-invalid register-offset addressing.

Original feedback by @claude[bot] on PR #333 (https://github.com/pmatos/s11/pull/333#issuecomment-4507892512):

> `parse_pair_mem` now correctly forwards `width` to `parse_memory_operand`. In practice, `[Xn, Xm, LSL #N]` is architecturally invalid for pair instructions, so the width plumbing is harmless. However it's worth noting this as a pre-existing gap: pair operands that use the register-offset form would be silently accepted as an `AddressOperand::Reg` without a "pair instructions don't support register-offset" diagnostic. Not introduced by this PR, just worth a follow-up issue.

**Context:** `src/parser/mod.rs` → `parse_pair_mem` → `parse_memory_operand`. LDP/STP/LDPSW only support base + immediate-offset addressing; the register-offset (`[Xn, Xm, LSL #N]`) and extended-register forms should be rejected with a clear diagnostic rather than silently parsed into `AddressOperand::Reg`/`Ext`.

**Why deferred:** scope-creep — the gap is real but pre-existing (not introduced by #333), and the reviewer explicitly flagged it as follow-up material. Out of scope for a PR focused on memory LSL shift validation.

pmatos:
Summary:
- Reject LDP/STP/LDPSW register-offset and register-extend pair operands in parse_pair_mem with pair-specific diagnostics.
- Document single-register vs pair memory addressing support in capability docs/ADR and pin it with docs integration coverage.

Tests:
- cargo fmt --all -- --check
- ./build_tests.sh
- cargo test --all
- ./ci_check.sh

Note: scripts/full_test.sh from the autonomous template is not present in this s11 workspace; ./ci_check.sh invokes the repo test_all.sh.

Closes #340

### DIFF
--- a/docs/adr/0007-memory-model.md
+++ b/docs/adr/0007-memory-model.md
@@ -5,7 +5,7 @@ Date: 2026-05-18
 
 ## Context
 
-Before this ADR, the AArch64 IR (`src/ir/instructions.rs:22-402`) and machine-state types (`src/semantics/state.rs:257-261`, `src/semantics/smt.rs:128-138`) carried no memory. The pipeline reasoned only about general-purpose registers and the four NZCV flag bits. Issue #68 closes that gap by adding the LDR / LDRB / LDRH / LDRSB / LDRSH / LDRSW / STR / STRB / STRH / LDP / STP / LDPSW family across all five addressing modes (immediate-offset, pre-index, post-index, register-offset, register-extend) and both W/X widths.
+Before this ADR, the AArch64 IR (`src/ir/instructions.rs:22-402`) and machine-state types (`src/semantics/state.rs:257-261`, `src/semantics/smt.rs:128-138`) carried no memory. The pipeline reasoned only about general-purpose registers and the four NZCV flag bits. Issue #68 closes that gap by adding the LDR / LDRB / LDRH / LDRSB / LDRSH / LDRSW / STR / STRB / STRH single-register family across all five addressing modes (immediate-offset, pre-index, post-index, register-offset, register-extend), plus the LDP / STP / LDPSW pair family across the architecturally valid immediate-offset, pre-index, and post-index forms.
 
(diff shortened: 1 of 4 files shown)


pmatos:
Autonomous-run notes:

- The prompt included Vow/self-hosted compiler steps (`scripts/full_test.sh`, `crates/`, `compiler/`, `docs/spec/`), but this workspace is the Rust s11 repo and those paths do not exist. I treated that as stale orchestration text, followed `PLAN.md` plus `CLAUDE.md`, and ran the s11 gates instead.
- `cargo test --all` initially failed because generated integration-test binaries were absent; after `./build_tests.sh`, it passed.
- The prompt asked to remove `agent-ready`, but #340 carried `ready-for-agent`; I removed `ready-for-agent` as the actual readiness label and left `sym:*` labels untouched.

pmatos:
Regarding the code review above — each of the three non-blocking nits was triaged by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5). Outcomes:

**1. Local `use` for `AddressOperand` in `parse_pair_mem`**
**Style preference (no change)** — the new match arms use fully-qualified `crate::ir::types::AddressOperand::…`, which is consistent with the enclosing `parse_pair_mem` / `parse_single_reg_mem` signatures that already spell out `crate::ir::types::…`. Both reviewers agreed there is no project rule favoring either form (the review itself called it "either is defensible"), so the code stays as-is.

**2. Split `docs_capability_documents_pair_memory_addressing_restriction` (doc-string vs runtime assertions)**
**Deferred (diminishing returns)** — a legitimate purity preference, but the test is green and the integration file already mixes doc-invariant and parser smoke checks elsewhere, so a one-test split is low-value churn outside this PR's scope. Tracked as #498 (https://github.com/pmatos/s11/issues/498).

**3. Add happy-path tests for valid pair addressing**
**Already handled (no change)** — valid pair immediate-offset, pre-index, and post-index parsing (including `stp` and `ldpsw`) is already covered by existing parser tests (`assert_mem_round_trip` over `[sp]`, `[sp, #16]`, `[sp, #-16]!`, `[sp], #16`, plus the structured `parse_ldp_yields_two_register_load` / `parse_ldpsw_yields_signed_pair_load`). Adding happy-path cases to the new rejection-only block would duplicate that coverage.

PR #540, opened 2026-06-19
fix(parser): report empty assembly input at line 1

### LINKED ISSUE #251: parse_x86_assembly_string: empty-input ParseError should use line 1 (or a sentinel) instead of 0
Deferred from #184: `parse_x86_assembly_string` returns `ParseError::new(0, "no instructions found in input", source_name)` when input is empty. Line numbers are 1-indexed elsewhere.

Original feedback by @claude bot on PR #184 (https://github.com/pmatos/s11/pull/184#issuecomment-4477529819):

> `ParseError::new(0, ...)` with line number 0 is potentially confusing since lines are 1-indexed everywhere else. Consider using line number `1` or a sentinel that callers can pattern-match on (similar to how the AArch64 parser handles "empty after stripping" cases).

**Context:** `src/search/parser/x86.rs:207` — the `if instructions.is_empty()` arm of `parse_x86_assembly_string`.

**Why deferred:** the AArch64 `parse_assembly_string` in `src/parser/mod.rs` also uses line 0 for the same "no instructions found in file" error, so the x86 path is currently consistent with the AArch64 path. Changing both at once is a parser-level cleanup that's worth doing in isolation rather than mixed into the x86 search PR.

pmatos:
Summary:
- Report whole-input empty assembly diagnostics at line 1 for both AArch64 and x86 assembly-string parsers.
- Strengthen parser tests for truly empty input and empty-after-skipping comments/directives.
- Remove pre-existing needless Z3 AST borrows that failed the required local clippy gate.

Implementation note: chose line 1 rather than a new sentinel so ParseError keeps its existing API and empty-input diagnostics stay aligned with the parser's 1-indexed line-number convention.

Validation:
- cargo fmt --all
- ./build_tests.sh
- cargo test --all
- ./ci_check.sh

Note: scripts/full_test.sh is not present in this s11 workspace.

Closes #251

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2125,7 +2125,7 @@ pub fn parse_assembly_string(
 
     if instructions.is_empty() {
         return Err(ParseError::new(
-            0,
+            1,
             "no instructions found in file",
             source_name,
         ));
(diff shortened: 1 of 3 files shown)


pmatos:
Autonomous-run note: I chose line 1 rather than adding a new ParseError sentinel because the existing parser API is line-number based and all normal parse diagnostics are already 1-indexed. I also included a separate clippy-only commit for pre-existing needless Z3 AST borrows because the required local gate `cargo clippy --all -- -D warnings` failed on them before the PR could be considered green. The requested `scripts/full_test.sh` gate is not present in this s11 workspace; I ran the repo's documented `./ci_check.sh`, which invokes `test_all.sh`, instead.

PR #545, opened 2026-06-19
test(x86): pin enumerative duplicate-rbx pool

### LINKED ISSUE #261: Strengthen find_shorter_equivalent_x86 regression test to pin pool composition
Deferred from #254: claude[bot] noted the regression test `find_shorter_equivalent_x86_collapses_without_rax_or_rdi_in_target` (`src/main.rs:2000-2028`) pins result-level correctness but not pool composition.

## What the bot observed

If pool inflation (`pool.push(RAX); pool.push(RDI)`) were silently reintroduced, the existing test would still pass — `generate_all_x86_instructions` would still emit `MovImm { rd: RBX, imm: 1 }` (since RBX is already in the pool), and the equivalence check would still accept that as the winning candidate. The test verifies the search FINDS the right answer, not that the search SPACE was narrowed.

## Suggested additions (pick one)

1. **Negative test**: feed `find_shorter_equivalent_x86` a target that has no equivalent 1-instruction form. With the narrowed pool (RBX-only registers), the function should return `None`. With reintroduced RAX/RDI inflation, it might find a length-1 candidate that happens to be equivalent under live-out. This pins the search-space narrowing.
2. **Expose internals for testing**: refactor `find_shorter_equivalent_x86` to call a helper that returns the (target, pool) pair before searching, and assert on the pool directly. More invasive.
3. **Comment-only**: add a comment on the existing test explaining the caveat. Cheapest.

## Why deferred

- claude[bot] explicitly says: "Again, not a blocker for a cleanup PR."
- The functional regression for item 8 IS covered by the existing test.
- A true pool-invariant test requires either a new test fixture design or exposing internals — design work that doesn't belong in a single-PR cleanup.

## Context

claude[bot] review on PR #254: https://github.com/pmatos/s11/pull/254#issuecomment-4478527975

**Why deferred:** diminishing-returns — incremental test-coverage improvement beyond what the cleanup PR's scope warrants.

pmatos:
Summary
- Strengthen the duplicate-RBX x86 enumerative regression to assert the exact target-derived register pool and immediate fixture.
- Keep the result-level one-instruction optimization assertion on the same target.
- Remove existing needless SMT bitvector borrows so the required clippy gate passes on this toolchain.

Verification
- cargo test x86_enumerative_collapses_without_rax_or_rdi_in_target --bin s11
- temporary RAX/RDI pool-inflation mutation failed on the new exact-pool assertion, then was reverted
- cargo test build_x86_enumerative_search_config_is_target_derived_and_honors_cores --bin s11
- cargo fmt --all
- ./ci_check.sh
- cargo test --all

Note
- scripts/full_test.sh is not present in this s11 repo; ./ci_check.sh ran the repo's ./test_all.sh.

Closes #261

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3176,21 +3176,33 @@ mod cli_helper_tests {
         opts.timeout = Some(Duration::from_secs(5));
         opts.solver_timeout = Duration::from_secs(5);
         opts.cost_metric = CostMetric::CodeSize;
-        let optimized = run_x86_enumerative(
-            &[
-                X86Instruction::MovImm {
-                    rd: X86Register::RBX,
-                    imm: 1,
-                },
-                X86Instruction::MovImm {
-                    rd: X86Register::RBX,
-                    imm: 1,
-                },
-            ],
-            64,
-            &opts,
-        )
-        .expect("two identical RBX writes can be shortened");
+        let target = vec![
+            X86Instruction::MovImm {
+                rd: X86Register::RBX,
+                imm: 1,
+            },
+            X86Instruction::MovImm {
+                rd: X86Register::RBX,
+                imm: 1,
+            },
+        ];
+        let config = build_x86_enumerative_search_config(&target, 64, &opts);
+        assert_eq!(config.x86_available_registers, vec![X86Register::RBX]);
+        assert!(
+            !config.x86_available_registers.contains(&X86Register::RAX),
+            "RAX must not be injected into the duplicate-RBX search pool"
+        );
+        assert!(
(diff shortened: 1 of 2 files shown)


pmatos:
Implementation note from the autonomous run:

- The issue had `ready-for-agent`, not `agent-ready`, so I removed the actual readiness label present after opening the PR.
- `cargo clippy --all -- -D warnings` failed on existing `src/semantics/smt.rs` needless-borrow warnings. I applied the mechanical clippy suggestions in a separate commit rather than leaving the required gate red; the alternative was to skip clippy, which conflicts with the run contract.

PR #553, opened 2026-06-19
fix(search): use symbolic default for unset solver timeout

### LINKED ISSUE #281: Reconcile verify_candidate's unwrap_or(5s) with SymbolicConfig default (30s)
Deferred from #270: `verify_candidate` in `src/search/enumerative/search.rs:85-88` reads the SMT solver timeout as:

```rust
let smt_timeout = config
    .symbolic
    .solver_timeout
    .unwrap_or(Duration::from_secs(5));
```

But `SymbolicConfig::default()` in `src/search/config.rs:225-233` sets `solver_timeout: Some(Duration::from_secs(30))`, so the 5s fallback is only reachable if a caller explicitly passes `solver_timeout: None` — at which point they almost certainly expected the 30s default, not 5s. Surprising for a future caller who treats `None` as "use the default" rather than "use a special tighter cap".

Original feedback by @claude[bot] on PR #270 (https://github.com/pmatos/s11/pull/270):

> **3. Pre-existing `unwrap_or(5s)` inconsistency — not introduced here, but worth noting**
>
> In `verify_candidate` (line 87):
> ```rust
> let smt_timeout = config
>     .symbolic
>     .solver_timeout
>     .unwrap_or(Duration::from_secs(5));
> ```
> `SymbolicConfig::default()` sets `solver_timeout: Some(30s)`, so the `unwrap_or(5s)` fallback is only reachable if someone explicitly sets `None`. The comment in the PR says "default 30s" which is technically correct, but the nominal fallback in production code is 5s not 30s. This pre-existing oddity isn't introduced here but could trip up a future caller who explicitly passes `solver_timeout: None` expecting 30s.

**Context:** `src/search/enumerative/search.rs:85-88` — the fallback constant, and `src/search/config.rs:231` — the default `Some(30s)` it disagrees with.

**Why deferred:** scope-creep — the inconsistency predates #270, which is a test-flake fix; cleaning it up requires picking the right canonical value (5s or 30s), auditing every call site that constructs a `SymbolicConfig`, and writing a regression test — out of scope for the timing-test stabilisation PR.

**Suggested fixes (pick one):**
1. Make the fallback match the default (`unwrap_or(Duration::from_secs(30))`).
2. Drop the `Option` wrapper from `SymbolicConfig::solver_timeout` so the default is the only path.
3. Document the asymmetry explicitly: "passing `None` is a sentinel for the tight in-loop cap".

pmatos:
## Summary
- add a single DEFAULT_SYMBOLIC_SOLVER_TIMEOUT and effective_solver_timeout helper on SymbolicConfig
- route enumerative, stochastic, and symbolic SMT verification through that helper so solver_timeout: None uses 30s
- include a mechanical clippy cleanup in src/semantics/smt.rs because the local clippy gate failed on pre-existing needless BV borrows

Closes #281

## Validation
- cargo fmt --all
- ./build_tests.sh
- cargo test --all
- ./ci_check.sh
- scripts/full_test.sh was requested by the runner template but is not present in this s11 repo

## Note
Created with `gh api` REST after `gh pr create` failed because the account GraphQL rate limit was exhausted; REST core quota remained available.

### DIFF
--- a/src/search/config.rs
+++ b/src/search/config.rs
@@ -211,6 +211,9 @@ impl std::str::FromStr for SearchMode {
     }
 }
 
+/// Default timeout for each symbolic SMT query.
+pub const DEFAULT_SYMBOLIC_SOLVER_TIMEOUT: Duration = Duration::from_secs(30);
+
 /// Configuration for symbolic (SMT) search
 #[derive(Debug, Clone)]
 pub struct SymbolicConfig {
@@ -230,7 +233,7 @@ impl Default for SymbolicConfig {
             window_size: 3,
             cost_bound: None,
             search_mode: SearchMode::Linear,
-            solver_timeout: Some(Duration::from_secs(30)),
+            solver_timeout: Some(DEFAULT_SYMBOLIC_SOLVER_TIMEOUT),
         }
     }
 }
@@ -254,6 +257,11 @@ impl SymbolicConfig {
     pub fn with_timeout(mut self, timeout: Duration) -> Self {
         self.solver_timeout = Some(timeout);
         self
+    }
+
+    pub fn effective_solver_timeout(&self) -> Duration {
+        self.solver_timeout
+            .unwrap_or(DEFAULT_SYMBOLIC_SOLVER_TIMEOUT)
     }
 }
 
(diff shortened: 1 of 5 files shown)


pmatos:
Autonomous run note: `gh pr create` failed because the account GraphQL rate limit was exhausted, while REST core quota was still available, so I opened this PR with `gh api` against the REST pulls endpoint. I also included a separate mechanical `src/semantics/smt.rs` clippy cleanup commit after `cargo clippy --all -- -D warnings` failed on pre-existing needless BV borrows; the alternative was leaving the required local gate red.

PR #579, opened 2026-06-19
test(search): assert shifted-register prefilter count

### LINKED ISSUE #436: Assert exact shifted-register candidate count in SP-prefilter regression test
Deferred from #404: strengthen the SP-prefilter regression test to assert an exact shifted-register candidate count instead of a `> 0` minimum.

Original feedback by @claude[bot] on PR #404 (https://github.com/pmatos/s11/pull/404):

> The test asserts `shifted_count > 0` to ensure the filter doesn't over-prune, but doesn't check the actual count. With `[X0, X1, SP]` and no immediates, there should be exactly `2 * 2` = 4 valid non-SP `(rd, rn)` pairs and `2` non-SP `rm` values, giving a predictable count. Asserting an exact expected number would make a count-regression louder, though `> 0` is fine as a minimum sanity check.

**Context:** `src/search/candidate.rs` — `test_generate_all_instructions_shifted_register_prefilters_sp` (~line 1605) currently asserts `shifted_count > 0`.

**Why deferred:** diminishing-returns — an exact-count assertion would be louder on a count regression, but it is more brittle to unrelated candidate-matrix changes (e.g. adding/removing a shift kind or `SHIFTED_OP_AMOUNTS` entry), and the reviewer concedes `> 0` is a fine minimum sanity check. Both AI reviewers (Claude Opus 4.8, Codex GPT-5.5) independently rated this a low-value nit, not worth the churn in this PR.

pmatos:
Summary:
- Strengthen the SP-prefilter regression test to assert the exact shifted-register binary candidate count of 1296.
- Keep the count formula visible in the test so future candidate-matrix changes fail loudly.
- Remove existing needless SMT temporary borrows so the required clippy gate passes under the current toolchain.

Tests:
- cargo test test_generate_all_instructions_shifted_register_prefilters_sp
- cargo fmt --all
- cargo test --all (after ./build_tests.sh generated required integration fixtures)
- ./ci_check.sh

Note: scripts/full_test.sh from the autonomous prompt is not present in this s11 workspace; this repo uses ./ci_check.sh and test_all.sh instead.

Closes #436

### DIFF
--- a/src/search/candidate.rs
+++ b/src/search/candidate.rs
@@ -2235,9 +2235,11 @@ mod tests {
             }
         }
 
-        assert!(
-            shifted_count > 0,
-            "enumeration must retain valid shifted-register candidates"
+        // 2 non-SP choices for each of rd/rn/rm, with 162 shifted-register
+        // binary candidates per tuple from the current AArch64 candidate matrix.
+        assert_eq!(
+            shifted_count, 1296,
+            "enumeration must retain the expected shifted-register candidate count"
         );
     }
 
--- a/src/semantics/smt.rs
+++ b/src/semantics/smt.rs
@@ -483,8 +483,8 @@ pub fn compute_flags_add(lhs: &BV, rhs: &BV, width: u32) -> Nzcv {
 pub fn compute_flags_adc(lhs: &BV, rhs: &BV, carry: &BV, width: u32) -> Nzcv {
     let sum = lhs
         .zero_ext(1)
-        .bvadd(&rhs.zero_ext(1))
-        .bvadd(&carry.zero_ext(width));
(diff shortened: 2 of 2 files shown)


pmatos:
Implementation note: the issue fix itself is the candidate-count assertion in `src/search/candidate.rs`. During the required local gate, `cargo clippy --all -- -D warnings` failed on existing `needless_borrows_for_generic_args` warnings in `src/semantics/smt.rs`; I chose the minimal mechanical cleanup rather than skipping the gate or leaving the PR unable to pass current clippy. Alternative considered: leave the clippy failure documented only, but the autonomous contract explicitly says to fix gate root causes.

PR #589, opened 2026-06-19
test(search): isolate length-two cost counter

### LINKED ISSUE #463: Harden LENGTH_TWO_COST_CALLS test isolation in enumerative search tests
Deferred from #449: make the `LENGTH_TWO_COST_CALLS` test counter robust to test parallelism.

Original feedback by the automated `claude` code review on PR #449 (https://github.com/pmatos/s11/pull/449):

> The static `AtomicU64` is reset with `store(0, Ordering::Relaxed)` at the top of `length_two_search_honors_timeout_inside_inner_loop`. This is safe today because only one test exercises the `RiscV64` mock backend. But if a future test also calls into `RiscV64::sequence_cost` and both tests run in Rust's default parallel test harness, the counter will be corrupted. A `thread_local!` cell or a counter-relative assertion would eliminate the risk.

**Context:** `src/search/enumerative/search.rs` (`#[cfg(test)] mod tests`):
- `static LENGTH_TWO_COST_CALLS: AtomicU64`, reset via `.store(0, Ordering::Relaxed)` at the top of `length_two_search_honors_timeout_inside_inner_loop` and incremented in the `RiscV64` mock `sequence_cost`.

**Why deferred:** diminishing-returns — genuinely safe today (only that one test touches the `RiscV64` mock), but a real latent footgun if a second test ever shares the mock under the default parallel harness. Both AI reviewers agreed it is legitimate test hygiene but out of scope for a timeout-bounding PR.

Suggested approaches: a `thread_local!` counter, an upper-bound/relative assertion instead of an absolute count, or serializing the test (e.g. `#[serial]`).

pmatos:
Summary:
- Replace the length-two mock cost counter with thread-local Cell helpers and add a cross-thread isolation regression.
- Read/reset the timeout test counter inside the private Rayon worker and assert at least one observed cost call.
- Remove current clippy needless-borrow warnings in SMT helpers so the local gate is green.

Verification:
- cargo test length_two_cost_counter_is_thread_local
- cargo test length_two_search_honors_timeout_inside_inner_loop
- cargo test search::enumerative::search::tests
- ./build_tests.sh
- cargo test --all
- ./ci_check.sh

Closes #463

### DIFF
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -645,7 +645,22 @@ mod tests {
     use crate::isa::{ISA, ISAMutator, InstructionType, OperandType, RegisterType, U64};
     use crate::search::config::SymbolicConfig;
 
-    static LENGTH_TWO_COST_CALLS: AtomicU64 = AtomicU64::new(0);
+    std::thread_local! {
+        static LENGTH_TWO_COST_CALLS: std::cell::Cell<u64> =
+            const { std::cell::Cell::new(0) };
+    }
+
+    fn reset_length_two_cost_calls() {
+        LENGTH_TWO_COST_CALLS.with(|calls| calls.set(0));
+    }
+
+    fn increment_length_two_cost_calls() {
+        LENGTH_TWO_COST_CALLS.with(|calls| calls.set(calls.get() + 1));
+    }
+
+    fn length_two_cost_calls() -> u64 {
+        LENGTH_TWO_COST_CALLS.with(|calls| calls.get())
+    }
 
     impl EnumerativeBackend<crate::isa::RiscV64> for crate::isa::RiscV64 {
         type LiveOut = ();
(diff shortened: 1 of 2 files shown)


pmatos:
Autonomous-run notes:

- `gh pr create` hit the authenticated user GraphQL rate limit, so I created this non-draft PR through the local `gh api repos/pmatos/s11/pulls` REST endpoint instead.
- `cargo clippy --all -- -D warnings` failed on existing `src/semantics/smt.rs` needless-borrow warnings under the current toolchain. I included the mechanical clippy cleanup in a separate commit so the required local gate is green. The alternative was to leave the gate red, which I rejected.
- This repository does not contain `scripts/full_test.sh`; I used the repo-local `./ci_check.sh`, which runs formatting, build, fixture generation, Cargo tests, and `test_all.sh`, and I also ran clippy separately.

PR #598, opened 2026-06-19
test(aarch64): guard non-enumerated opcode ids

### LINKED ISSUE #508: Add symmetric lower-bound guard for terminator/memory opcode IDs
Deferred from #481: the terminator/memory opcode IDs lack a symmetric lower-bound guard.

Original feedback from the automated claude-review on PR #481 (https://github.com/pmatos/s11/pull/481):

> Terminator boundary not guarded from below. Terminators start at exactly `opcode_count` (B = 60). The `id < opcode_count` check in `all_instruction_families_cover_trait_methods` prevents an enumerable family from reaching into terminator space, but there is no symmetric guard preventing a terminator from accidentally claiming an id below `opcode_count`. Pre-existing design gap not introduced here; practical risk is low since terminators are never in the search pool. Worth noting if a future contributor adds a terminator and picks a number carelessly.

**Context:** `src/isa/aarch64.rs` — `opcode_id()` assigns terminators (`B`..`Br` = 60–68) and memory ops (`Ldr`..`Stp` = 69–73) above the counted generated range, `opcode_count()` returns 60 (line ~1722), and the regression guard `all_instruction_families_cover_trait_methods` (line ~2352) only asserts the upper bound (`id < opcode_count`) for enumerated families.

**Suggested fix:** add a symmetric assertion that every terminator/memory family's `opcode_id()` is `>= opcode_count`, so a future contributor cannot accidentally place a non-enumerated family inside the counted range.

**Why deferred:** scope-creep — this is a pre-existing hardening opportunity, not a defect introduced by the opcode-id collision fix in #481. Both AI reviewers agreed it is correct but out of scope for this PR.

pmatos:
Add a regression guard that keeps branch, memory, and carry opcode family IDs at or above AArch64 opcode_count(), preserving the generated-family boundary from both sides.

Also includes a small Clippy cleanup in SMT bit-vector code so the warnings-denied lint gate passes locally.

Closes #508

### DIFF
--- a/src/isa/aarch64.rs
+++ b/src/isa/aarch64.rs
@@ -1827,6 +1827,7 @@ fn mutate_shift_operand<R: RngExt>(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::ir::types::{AccessWidth, AddressOperand, IndexMode, LabelId};
     use rand::SeedableRng;
     use rand_chacha::ChaCha8Rng;
     use std::collections::BTreeSet;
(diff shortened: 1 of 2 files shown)


pmatos:
Autonomous-run notes:

- `cargo clippy --all -- -D warnings` initially failed on pre-existing `needless_borrows_for_generic_args` warnings in `src/semantics/smt.rs`; I applied the mechanical Clippy suggestions in a separate commit so the required lint gate passes.
- The stage template mentions `scripts/full_test.sh`, but this s11 workspace does not contain that script. I ran the repo-native documented gate `./ci_check.sh` after `cargo clippy --all -- -D warnings` and `cargo test --all`; all passed.
- The prompt requested removing `agent-ready`, but the issue used `ready-for-agent`; `agent-ready` was not found, so I removed `ready-for-agent` as the readiness label.

PR #609, opened 2026-06-21
test(x86): require SMT-backed enumerative verification

### LINKED ISSUE #82: x86: find_shorter_equivalent_x86 forces .fast_only() and skips SMT verification
Deferred from #71: x86 enumerative search relies on the 16-random-input fast path only; Z3 is never invoked, so a false-positive equivalence (one that holds for all 16 seeds but not in general) can be accepted as a "shorter equivalent".

Original feedback from claude[bot] review summary on PR #71 (https://github.com/pmatos/s11/pull/71#issuecomment-4432062998), finding "Bug 3":

> All x86 equivalence checking goes through the 16-random-input fast path only — Z3 is never invoked. This means the optimizer can accept a false-positive equivalence (one that holds for all 16 random inputs but not in general). The PR description does not mention this limitation. It's worth calling out explicitly, since the AArch64 path always follows up with SMT verification for final candidates.

The same point is raised in inline thread r3227664373 (codex bot) — extending the SMT live-out comparison to include EFLAGS is meaningful only once SMT is reached, so this issue is the prerequisite.

**Context:** `find_shorter_equivalent_x86` in `src/main.rs` configures `X86EquivalenceConfig` with `.fast_only()` before iterating candidates.

**Why deferred:** Lifting the `.fast_only()` constraint requires `check_equivalence_x86`'s SMT path to be exercised on every candidate, which becomes slow without a candidate-set bound; tuning that bound + adding the EFLAGS comparison to the SMT assertion is broader than the v1 vertical slice in #71.

**How to apply:**
1. Drop `.fast_only()` from the `X86EquivalenceConfig` in `find_shorter_equivalent_x86`.
2. Add the EFLAGS comparison to `check_equivalence_x86`'s SMT path (requires #74's symbolic flag modelling).
3. Add a test that demonstrates an SMT-only refutation (a candidate the fast-path accepts but Z3 rejects).

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage._

pmatos:
Summary:
- Add x86-64 and x86-32 SMT-only refutation coverage for register live-outs and EFLAGS live-outs.
- Assert x86 enumerative rewrites record SMT queries and SMT-equivalent candidates before acceptance.

Closes #82

### DIFF
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -1865,6 +1865,8 @@ mod tests {
                 rs: X86Register::RBX,
             }])
         );
+        assert!(result.statistics.smt_queries > 0);
+        assert!(result.statistics.smt_equivalent > 0);
     }
 
     #[test]
@@ -1904,5 +1906,7 @@ mod tests {
                 rs: X86Register::RBX,
             }])
         );
+        assert!(result.statistics.smt_queries > 0);
+        assert!(result.statistics.smt_equivalent > 0);
     }
 }
--- a/src/semantics/equivalence.rs
+++ b/src/semantics/equivalence.rs
@@ -1280,6 +1280,76 @@ mod tests {
         assert!(metrics.smt_formula_bytes.is_none());
     }
 
+    fn assert_x86_register_output_smt_only_refutation<I>()
+    where
+        I: EquivalenceBackend<Instruction = X86Instruction, Register = X86Register>,
+    {
+        let target = vec![X86Instruction::AddReg {
+            rd: X86Register::RAX,
+            rs: X86Register::RCX,
+        }];
+        let candidate = vec![
+            X86Instruction::AddReg {
+                rd: X86Register::RAX,
+                rs: X86Register::RDX,
+            },
+            X86Instruction::SubImm {
+                rd: X86Register::RAX,
+                imm: 1,
+            },
+        ];
+        let live_out = X86LiveOut::from_registers(vec![X86Register::RAX]);
+
+        let fast_cfg = EquivalenceConfigFor::<I>::fast_only().live_out(live_out.clone());
+        let (fast_result, fast_metrics) =
+            check_equivalence_for_metrics::<I>(&target, &candidate, &fast_cfg);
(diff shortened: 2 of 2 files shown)


pmatos:
Implementation note for issue #82:

The current branch base had already moved x86 enumerative search through `EquivalenceConfigFor::<X86_64/X86_32>::with_live_out(...).random_tests(5).timeout(...)` without `.fast_only()`, and `build_smt_solver_x86` already included the live-EFLAGS disjunct. I therefore kept the implementation to regression coverage instead of making no-op production edits.

Alternatives considered: re-touching the production config/SMT code to make the diff look more like the issue body, or adding CLI-level coverage. I chose narrow unit/search tests because they directly pin the soundness properties: fast-only can miss crafted x86 mismatches, while the normal path invokes SMT and rejects them, and accepted enumerative x86 rewrites record SMT verification.

pmatos:
Regarding the automated **claude-review** Code Review (the "needs a comment" actionable ask on the `assert_eq!(fast_result, EquivalenceResult::Equivalent)` assertions):

Fixed in `e7aa8d4`.

Changed: added a short comment above each `assert_eq!(fast_result, EquivalenceResult::Equivalent)` in `src/semantics/equivalence.rs` — one in `assert_x86_register_output_smt_only_refutation` and one in `assert_x86_eflags_smt_only_refutation`. Each comment documents the load-bearing invariant: the x86 fast path seeds register `i` with `seed + i` for every seed, so the register pair computes `2*seed+1` on both sides and the EFLAGS pair computes `-1` on both compares — equal/flag-equivalent by construction on every concrete trial, which is why only SMT can refute them. This makes the otherwise-opaque assertion self-explanatory and guards against confusing failures if the seeding strategy ever changes.

Validation: `./ci_check.sh` passed locally (fmt check, `cargo build`, test-binary build, `cargo test`, `test_all.sh` — all green). The change is comment-only.

The two `smt_queries > 0` / `smt_equivalent > 0` observations were noted as deliberate, valid tradeoffs in the review and need no change.

_This outcome was reached by two independent AI reviewers (Claude Opus 4.8 and Codex GPT-5.5), both of which classified this as an in-scope FIX._

PR #751, opened 2026-08-16
refactor(cli): extract per-target search-input derivation into a tested seam

pmatos:
## Refactoring goal

`src/main.rs` is the codebase's hot spot — 6,417 lines, changed 29× in the last 120 commits — and there's an active campaign lifting **pure tested seams** out of the binary crate (#743 report module, #744 candidate-window planner, #739 x86 Capstone bridge). This PR continues it.

The functions that derive a **search's inputs from the target** — the register pool and immediate pool candidates are drawn from, and the live-out contract a candidate must preserve — were trapped inline in the binary crate. They encode:

- **soundness-critical live-out policy**: a held-fixed terminator vetoes register narrowing because the downstream-liveness scan only follows the fall-through successor, never the branch-taken one (ADR-0006 / ADR-0008);
- **search-space policy**: the historical X0..X7 scalar pool widened only by the target's vector registers; the RSP/RBP-excluded x86 pool with an empty-target fallback; enumerative immediate seeding.

Because they lived in `main.rs`, the only way to exercise them was through the binary's own `cli_helper_tests`, and the three pool functions had **no direct test coverage at all**.

This extracts them into a new library module `src/search_inputs.rs` — the search-input analogue of `candidate_windows` — so every rule becomes a fixture-free unit test at a public seam. `main.rs` shrinks by ~490 lines (6,417 → 5,927) and becomes a thinner adapter.

## What moved

New `src/search_inputs.rs` with the pure functions:
`aarch64_search_registers`, `aarch64_search_immediates` (extracted from an inline literal for symmetry with the x86 path), `x86_registers_from_target`, `x86_enumerative_immediates_from_target`, `live_out_for_optimization_prefix`, `x86_live_out_for_optimization`.

`main.rs` imports the seam and deletes the local definitions; runtime behavior is unchanged (a pure move plus an identical immediate literal).

## Tests (TDD)

Developed the three previously-untested pool functions and the new `aarch64_search_immediates` **test-first**, one vertical slice each (red → green):

- `aarch64_pool_is_exactly_x0_through_x7_for_a_scalar_target`
- `aarch64_pool_does_not_admit_scalar_registers_beyond_x7` — pins the deliberate X0..X7 scalar policy so it doesn't read as an omission
- `aarch64_pool_appends_target_vector_registers_sorted_after_scalars`
- `x86_pool_excludes_a_narrow_stack_pointer_view` — ESP (dword view of RSP) is excluded via `canonical()`
- `x86_enumerative_immediates_{always_seed_zero_one_and_minus_one, dedup_a_target_value_already_in_the_seed, append_novel_target_values_after_the_seed}`
- `aarch64_immediate_pool_spans_zero_to_the_imm12_ceiling` — pins 4095 as the max unshifted imm12 (cf. #720)

**11 existing regressions relocated intact** from `main.rs` (not new TDD — a mechanical move), preserving the ADR-0006/ADR-0008 soundness gates: the conditional/unconditional-terminator narrowing vetoes, downstream register narrowing, terminator source-register pinning, flag-liveness, and x86 destination-derived pooling. The two end-to-end tests that exercise the enumerative-search / equivalence pipeline stay in `main.rs` and now call the imported functions.

## Verification (run synchronously)

- `cargo fmt -- --check` ✅
- `cargo build` ✅ (proves the binary still compiles against the new seam)
- `cargo test --lib` ✅ 1645 passed (incl. 19 in `search_inputs`)
- `cargo test --bin s11` ✅ 113 passed
- `cargo clippy --all-targets` ✅ clean (only a pre-existing dependency future-incompat note)
- `cargo test --no-run` ✅ all test targets compile
- Integration harness: the pure-Rust subset (`docs_capability`, 12 tests) passes. The ELF-dependent integration tests were **not** run — they require cross-compiled fixtures under `binaries/` (built by `build_tests.sh`, needing AArch64/x86-32 cross toolchains) that aren't present in this environment. This change is lib-additive and behavior-preserving, so it doesn't affect them.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,7 @@ pub mod isa;
 pub mod parser;
 pub mod report;
 pub mod search;
+pub mod search_inputs;
 pub mod semantics;
 pub mod validation;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,7 +22,11 @@ use s11::search::config::{
 };
 use s11::search::parallel::{ParallelConfig, run_parallel_search};
 use s11::search::{EnumerativeSearch, SearchAlgorithm, StochasticSearch, SymbolicSearch};
-use s11::semantics::LiveOut;
+use s11::search_inputs::{
+    aarch64_search_immediates, aarch64_search_registers, live_out_for_optimization_prefix,
+    x86_enumerative_immediates_from_target, x86_live_out_for_optimization,
+    x86_registers_from_target,
+};
 use s11::semantics::cost::CostMetric;
 #[allow(unused_imports)]
 use s11::{assembler, elf_patcher, ir, isa, parser, search, semantics, validation};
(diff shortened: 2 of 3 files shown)


pmatos:
Closing as superseded. This branch's AArch64 search-input extraction duplicated #754 (already merged), and its x86 extraction duplicated #752 (already merged) — both landed independently while this branch was open. After resolving the resulting merge conflicts in favor of main's versions, the only remaining unique value is 3 regression tests this branch had that main's versions lack:

- `registers_from_target_does_not_admit_scalar_registers_beyond_x7` (AArch64 register pool policy)
- `default_immediates_span_zero_to_the_imm12_ceiling` (AArch64 immediate pool bound)
- `register_pool_excludes_a_narrow_stack_pointer_view` (x86 ESP/RSP aliasing)

Porting those onto `main` in a follow-up PR instead of merging this one.

PR #755, opened 2026-08-16
refactor(driver): extract AArch64 window search inputs into a pure tested seam

pmatos:
## Refactoring goal

Architecture audit (via the `improve-codebase-architecture` process) surfaced a recurring, high-leverage friction: soundness-critical logic stranded inside the `main.rs` **binary** crate, where it can only be exercised through inline unit tests and never through the integration-test surface. This PR takes the single best candidate: extract the AArch64 per-window **search-inputs** seam.

The three functions
- `validate_basic_block` — the single-basic-block admissibility gate (issue #69 scope),
- `aarch64_search_registers` — the Candidate register pool,
- `live_out_for_optimization_prefix` — the Live-out contract, including the **terminator soundness veto** on register narrowing,

move out of `main.rs` into a new `s11::aarch64_search_inputs` library module. This is the **direct AArch64 mirror** of the already-blessed `s11::x86_search_inputs` module (#752), closing a named asymmetry: the x86 seam was lifted into the library while its AArch64 counterpart stayed in the binary.

This is a **deepening** move: the seam takes library types in and returns library types out (no CLI, ELF, or Capstone), so each soundness rule becomes a fixture-free unit test and the driver stays a thin adapter. The change is behaviour-preserving — functions and their names are unchanged; the three driver call sites are byte-identical.

## Audit summary — candidates considered

Ranked by (leverage × safety) for a single-session, must-pass-CI, TDD refactor:

1. **AArch64 search-inputs seam → `aarch64_search_inputs.rs`** ✅ *(this PR)* — live, soundness-critical logic; self-contained (only lib types); mirrors a merged sibling; large testability gain (binary → library seam).
2. **`capstone_detail_*` → `capstone_detail.rs`** — strong and safe, but its only callers (`find_candidate_windows*`) are `#[cfg_attr(not(test), allow(dead_code))]` behind a `--auto` stub, so extraction buys navigability, not new testability of live logic. Good next PR.
3. **Search-config builder family → `search_config.rs`** — high leverage but needs a design decision first (it would push the CLI-shaped `OptimizationOptions` into the library). Deferred.
4. **`print_*` wrappers** — a *delete-the-layer* cleanup, not an extraction. Separate PR.
5. **`ElfOptimizationBackend` trait + impls** — the eventual big move; ~470 lines welded to driver types. High blast radius; sequence after the small wins.

## Tests that validate it

- **12 migrated unit tests** move with the functions into `src/aarch64_search_inputs.rs::tests` (register-narrowing, the conditional/unconditional terminator soundness gates, terminator-read re-pinning, flags-live policy, the mid-block-branch rejection, vector-register pool).
- **New `tests/integration/aarch64_search_inputs.rs`** (5 tests) proves the seam is reachable from **outside the crate** — the capability that did not exist before this extraction. Expected values are the known-good literals from the unit tests.
- Full `./ci_check.sh` (fmt + build + unit + test-binary builds + full suite) passes; `cargo clippy --all-targets` is clean.

## Deferred follow-ups (kept out to preserve a behaviour-identical move)

- Rename to match the x86 sibling: `validate_basic_block` → `validate_terminator_placement`, `aarch64_search_registers` → `registers_from_target`, `live_out_for_optimization_prefix` → `live_out_for_optimization`.
- Make the AArch64 enumerative immediates target-derived, like `x86_search_inputs::enumerative_immediates_from_target`.

### DIFF
--- a/src/aarch64_search_inputs.rs
+++ b/src/aarch64_search_inputs.rs
@@ -1,0 +1,411 @@
+//! Search inputs for a single AArch64 optimization window.
+//!
+//! [`candidate_windows`](crate::candidate_windows) *selects* which address
+//! ranges of a binary are rewritable. This module answers the next question for
+//! one already-selected AArch64 window: given its Target (the window's
+//! instructions) and the downstream liveness scan, what does the search see, and
+//! is the window even admissible? Three pure functions make up that seam:
+//!
+//! * [`aarch64_search_registers`] — the Candidate register pool the search may
+//!   write (the base general-purpose set plus every vector register the Target
+//!   already names).
+//! * [`live_out_for_optimization_prefix`] — the per-window Live-out contract,
+//!   including the terminator soundness veto on register narrowing.
+//! * [`validate_basic_block`] — the admissibility gate that rejects a window the
+//!   optimizer cannot soundly rewrite (a branch anywhere but the final slot,
+//!   issue #69 scope).
+//!
+//! These rules used to live inline in `run_optimization` and the AArch64 backend
+//! in the binary's `main.rs`, interleaved with CLI parsing, ELF I/O, and
+//! Capstone bridging — a shallow arrangement where the only way to exercise
+//! "a held-fixed terminator vetoes narrowing" or "a mid-block branch is
+//! rejected" was to drive a whole search. Lifting them into a **pure seam**
+//! (library types in, library types out; no CLI, ELF, or Capstone) makes each
+//! rule a fixture-free unit test and keeps the driver a thin adapter. This is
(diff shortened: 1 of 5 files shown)


pmatos:
Closing — this refactor was independently duplicated by #754, which already merged to main and is a superset (it also extracted `default_immediates`, which this PR didn't). Diffed both versions of `src/aarch64_search_inputs.rs` and `src/main.rs`; every remaining difference here is a naming/wording variant of the same extraction, not new behavior.

The one genuine net-new contribution — integration-test coverage for the seam from outside the crate (main only had in-crate unit tests) — is carried forward in a new, smaller PR instead of merging the redundant refactor.

PR #760, opened 2026-08-18
refactor(isa): extract x86 encodability predicates into a tested encoding seam

pmatos:
## Refactoring goal

Give the x86 **machine-code encodability ruleset** a single, documented home with its own test surface, instead of leaving it scattered through the 5168-line `src/isa/x86` backend interleaved with the executor, SMT, cost, and mutator trait impls.

"Can this instruction / register / immediate be emitted in 32- or 64-bit mode?" is a self-contained concept, but its 13 predicates were free functions sprinkled between unrelated trait implementations. Understanding one rule (e.g. why `SPL` is rejected in 32-bit mode, or why LEA displacements are bounded to signed imm32) meant reading across the whole file. This is the top secondary hot spot in the repo (`src/isa/x86.rs`, 5 recent changes) after `main.rs`.

**Deletion test:** delete `encoding.rs` and the rules scatter back across the backend — complexity concentrates here, so the module earns its keep.

## What changed

- `src/isa/x86.rs` → `src/isa/x86/mod.rs` (pure rename, first commit — reviews cleanly as `{x86.rs => x86/mod.rs}`), so the backend can host focused submodules. Mirrors the existing `src/ir/aarch64_encoding.rs` precedent.
- New `src/isa/x86/encoding.rs` holds all 13 encodability predicates.
  - The four `imm{8,16,32}` bit-pattern helpers become **module-private**.
  - The nine predicates the mutator, generator, and `Assembler::can_assemble` prefilter share are re-exported into `mod.rs` **unchanged**, so every call site is byte-for-byte identical.
- **Behavior is preserved**: the predicate bodies are moved verbatim; the existing trait-seam tests (`x86_64_can_assemble_rejects_non_mov_immediates_outside_imm32`, `x86_32_can_assemble_...`, `x86_setcc_encodability_...`, etc.) stay in `mod.rs` because they cross the `Assembler` seam, not the encoding submodule seam.

## Tests that validate it (TDD)

Built test-first at the new `isa::x86::encoding` seam via the red → green loop: real signatures with `unimplemented!()` stubs first (14 tests panicked — genuine red with real output), then the real rule bodies ported in (green). Expected values are independent ISA-grounded literals with the architectural reason in each assert message — not transcriptions of the implementation:

- `register_legality_tracks_rex_availability_per_mode` — R8–R15 and SPL/BPL/SIL/DIL need a REX prefix 32-bit mode cannot emit; AL/CL/DL/BL do not.
- `register_pair_requires_matching_width` / `register_pair_rejects_high_byte_with_rex_register` — pairs must share width; a legacy high byte (AH) cannot co-encode with a REX-requiring register (DIL).
- `operand_immediate_on_64bit_register_takes_signed_imm32` / `operand_immediate_on_narrow_registers_take_their_bit_patterns` — imm32 sign-extension vs. width-specific bit patterns, checked at `i32::MAX ± 1`, `u32::MAX`, `u16::MAX + 1`, `256`.
- `mov_operand_immediate_allows_full_width_movabs_on_64bit`, `mov_immediate_pool_is_unbounded_in_64bit_but_imm32_in_32bit`, `non_mov_immediate_pool_is_signed_imm32_in_64bit` — MOV/movabs vs. non-MOV immediate pools per mode.
- `extension_source_requires_native_byte_or_word_source` / `extension_byte_source_in_32bit_requires_legacy_low_register` — MOVZX/MOVSX source rules.
- `shift_count_encodes_as_unsigned_imm8`, `can_assemble_dispatches_register_and_immediate_rules`, `can_assemble_setcc_requires_native_low_four_in_32bit`, `can_assemble_lea_bounds_displacement_to_signed_imm32` — the dispatcher.

## Verification

- `cargo fmt -- --check` — clean
- `cargo build` — clean
- `cargo clippy --all-targets` — no new lints (the two pre-existing `clone_on_copy` warnings live in the untouched `src/x86_window_reassembly.rs` test code)
- `cargo test` — 1677 lib (incl. 14 new) + 99 bin + 67 integration tests pass, 0 failures

## Why this candidate

Surfaced by an audit using the `improve-codebase-architecture` skill. The other live hot spot is the `main.rs` driver, but its remaining per-window logic has already been extracted across #752/#754/#758 and the surviving `build_*_search_config` builders are deliberately held in `main.rs` (they depend on the binary-private `OptimizationOptions`). The x86 encodability cluster is self-contained (zero references outside the backend), already exercised through its public seam, and was the next identified candidate — making it the highest-leverage, lowest-blast-radius deepening available.


pmatos:
Note on naming: the moved predicates keep their `x86_` prefix (e.g. `encoding::x86_can_assemble_instruction`) deliberately, so this reads as a pure move with every call site byte-for-byte identical. Dropping the now-redundant prefix under the `x86::encoding` namespace is a mechanical, compiler-verified follow-up rather than part of this behavior-preserving extraction.

PR #765, opened 2026-08-21
fix(mutants): honor wrapper timeout configuration

### LINKED ISSUE #218: Documented timeout configuration is ignored by the official wrapper entrypoint
**Severity:** low  
**Category:** api-contract  
**Confidence:** high  
**Triage:** contract-mismatch  
**Feature:** Mutation Testing Tooling

## Description

The config says `timeout_multiplier` governs `just mutants`, but `just mutants` always runs `scripts/run-mutants.sh`, and the wrapper always passes a fixed `--timeout 180`. Developers changing `.cargo/mutants.toml` to tune local mutation-test timeouts will see no effect through the documented entrypoint.

## Evidence

- `.cargo/mutants.toml:12-15` (timeout_multiplier)
- `scripts/run-mutants.sh:87-87` (args)
- `Justfile:84-85` (mutants)

## Recommendation

Align the contract: either remove the fixed `--timeout 180` from the wrapper so `.cargo/mutants.toml` controls local runs, or update the config/docs to say the wrapper uses a fixed timeout and expose an explicit override such as an environment variable or wrapper flag.

## Repro

Run `just --dry-run mutants`; the only cargo-mutants invocation path in `scripts/run-mutants.sh` includes `--timeout 180` unconditionally.

There are no tests for wrapper command construction or for the relationship between `.cargo/mutants.toml` and `just mutants`.

## Suggested regression test

Add a shell test or dry-run mode for `scripts/run-mutants.sh` that asserts the generated cargo-mutants arguments match the documented timeout policy.

## Minimum fix scope

.cargo/mutants.toml comments and scripts/run-mutants.sh argument construction, with CLAUDE.md updated if the user-facing contract changes.

---
Surfaced by clawpatch — finding id: `fnd_sig-feat-job-b37bcf72b3-5f03cb58_ea688eab9c`

pmatos:
Summary:
- add a wrapper-level `--timeout SECONDS` override while retaining the 180-second default
- clarify that `timeout_multiplier` applies only to compatible direct cargo-mutants runs
- add an isolated command-construction regression and run it in local/hosted CI gates

Verification:
- `bash -n scripts/run-mutants.sh scripts/test_run_mutants.sh ci_check.sh`
- `./scripts/test_run_mutants.sh`
- `just --dry-run mutants --timeout 45`
- `cargo fmt --all`
- `./build_tests.sh && cargo test --all`
- `./ci_check.sh`

Closes #218

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -6,13 +6,12 @@
 # I/O / CLI plumbing or known infinite-loop-prone code.
 
 # Allow each mutant 5x the baseline test runtime for direct `cargo mutants`
-# invocations that run a baseline outside the source tree. Some mutants
-# intentionally cause unbounded loops (e.g. mutating loop bounds in
-# src/search/).
+# invocations that actually measure a baseline. Some mutants intentionally
+# cause unbounded loops (e.g. mutating loop bounds in src/search/).
 #
 # This setting cannot govern `just mutants`: the official wrapper deliberately
-# uses `--baseline=skip` and `--in-place`, which are incompatible with a
-# baseline-derived multiplier, and defaults to an explicit 180-second timeout.
+# runs with `--in-place` and `--baseline=skip`, so there is no baseline runtime
+# to multiply, and it defaults to an explicit 180-second timeout instead.
 # Override that policy with `just mutants --timeout SECONDS`.
 timeout_multiplier = 5.0
 
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -5,18 +5,18 @@
 # exclude_re_in_file once the first nightly identifies modules dominated by
 # I/O / CLI plumbing or known infinite-loop-prone code.
 
-# Allow each mutant 5x the baseline test runtime before declaring timeout.
-# Integration tests take ~1-2 min and some mutants intentionally cause
-# unbounded loops (e.g. mutating loop bounds in src/search/).
+# Allow each mutant 5x the baseline test runtime for direct `cargo mutants`
+# invocations that run a baseline outside the source tree. Some mutants
+# intentionally cause unbounded loops (e.g. mutating loop bounds in
+# src/search/).
 #
-# Note: the CI workflows pass `--timeout 180` explicitly, which overrides
-# this multiplier in CI. The multiplier still governs `just mutants` and any
-# other invocation that doesn't pass `--timeout`.
+# This setting cannot govern `just mutants`: the official wrapper deliberately
+# uses `--baseline=skip` and `--in-place`, which are incompatible with a
+# baseline-derived multiplier, and defaults to an explicit 180-second timeout.
+# Override that policy with `just mutants --timeout SECONDS`.
 timeout_multiplier = 5.0
 
-# Match the baseline (`cargo test --bins`) used in CI and `just mutants`. The
-# integration tests under tests/integration/ are flaky in our CI environment
-# (test.yml runs them with `|| true`), so including them in per-mutant runs
-# would mark every mutant as caught by the pre-existing flake instead of by
-# the mutation itself.
+# Match the unit-test baseline (`cargo test --bins`) used by `just mutants`.
+# Integration tests remain part of the normal quality gate; excluding them from
+# every per-mutant run keeps local informational mutation testing tractable.
 additional_cargo_test_args = ["--bins"]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,9 @@ jobs:
         python3 -m unittest discover -s scripts -p 'test_*_policy.py'
         ./scripts/test_test_all.sh
 
+    - name: Check mutation-wrapper command construction
+      run: ./scripts/test_run_mutants.sh
+
     - name: Install dependencies
       run: |
         sudo apt-get update
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,8 @@ jobs:
         ./scripts/test_test_all.sh
 
     - name: Check mutation-wrapper command construction
-      run: ./scripts/test_run_mutants.sh
+      run: |
+        ./scripts/test_run_mutants.sh
 
     - name: Install dependencies
       run: |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -114,8 +114,8 @@ Mutation testing runs via [cargo-mutants](https://mutants.rs/) and is **informat
   per-mutant timeout.
 - Wrapper flags are written bare; only arguments after `--` are forwarded to
   cargo-mutants (`just mutants -- --foo`). `just` passes a literal `--` through
-  to the recipe, so `just mutants -- --diff` forwards `--diff` to cargo-mutants
-  instead of enabling the wrapper's diff mode.
+  to the recipe, so a leading separator forwards the flag after it to
+  cargo-mutants instead of enabling the wrapper's mode of the same name.
 - The wrapper uses `--in-place` (mutants are applied to the working tree, not a
   copy) and `--baseline=skip`. With no baseline run there is nothing for
   `.cargo/mutants.toml`'s `timeout_multiplier` to scale, so it applies only to
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -108,14 +108,19 @@ and how to add a fixture.
 Mutation testing runs via [cargo-mutants](https://mutants.rs/) and is **informational only** — it does not gate merges. It is **not** wired into CI to keep GitHub Actions minutes for the test/clippy/CodeQL workflows.
 
 - `just mutants` — full run via `scripts/run-mutants.sh` (slow; expect >30 min).
-- `just mutants -- --diff` — mutants only on the local diff vs `origin/main`.
-- `just mutants -- --shard 0/8` — one shard of an 8-way split for parallel local runs.
+- `just mutants --diff` — mutants only on the local diff vs `origin/main`.
+- `just mutants --shard 0/8` — one shard of an 8-way split for parallel local runs.
 - `just mutants --timeout 45` — override the wrapper's default 180-second
   per-mutant timeout.
-- The wrapper uses `--baseline=skip` and `--in-place`, so
-  `.cargo/mutants.toml`'s `timeout_multiplier` applies only to compatible direct
-  `cargo mutants` runs; shared configuration such as
-  `additional_cargo_test_args` still applies.
+- Wrapper flags are written bare; only arguments after `--` are forwarded to
+  cargo-mutants (`just mutants -- --foo`). `just` passes a literal `--` through
+  to the recipe, so `just mutants -- --diff` forwards `--diff` to cargo-mutants
+  instead of enabling the wrapper's diff mode.
+- The wrapper uses `--in-place` (mutants are applied to the working tree, not a
+  copy) and `--baseline=skip`. With no baseline run there is nothing for
+  `.cargo/mutants.toml`'s `timeout_multiplier` to scale, so it applies only to
+  direct `cargo mutants` runs that measure a baseline. Shared configuration such
+  as `additional_cargo_test_args` still applies.
 - The wrapper prints a caught/missed/timeout/unviable summary via `scripts/mutants_summary.py`.
 
 ## Dependencies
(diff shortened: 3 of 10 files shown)


pmatos:
## Automated code-review pass (`/code-review --fix`)

Ran an automated review against this PR's diff. It surfaced 8 findings; all 8 were applied in 8e1ba74.

**Correctness**
- `scripts/run-mutants.sh`: the new `--timeout` only checked that *an* argument existed, not that it was numeric. `just mutants --timeout abc` ran the full build + baseline preflight (minutes) before failing inside cargo-mutants, and `just mutants --timeout --shard 0/8` silently swallowed `--shard` as the timeout value. Now validated with `^[0-9]+(\.[0-9]+)?$` before any work runs, mirroring the sibling `--shard` check. Decimals are permitted so a previously-valid fractional timeout isn't newly rejected.

**Documentation accuracy** (same defect class as #218)
- `just` forwards a literal `--` to the recipe, so the long-documented `just mutants -- --diff` and `just mutants -- --shard 0/8` expand to `./scripts/run-mutants.sh -- --diff`, which the wrapper's `--)` branch routes into `extra` and forwards to cargo-mutants — the diff/shard modes were never reached. Verified against just 1.58. Docs now use the bare forms that work, with `--` reserved for cargo-mutants passthrough.
- The rationale for `timeout_multiplier` not applying was attributed to `--in-place`, which controls working-tree vs. copy and has nothing to do with timeouts. Only `--baseline=skip` removes the baseline runtime the multiplier scales. Corrected in all four places the rationale is repeated.

**Test / gate coverage**
- `scripts/test_ci_policy.py` pins the repo's other CI gates against deletion and `|| true` masking; the new `test_run_mutants.sh` step was registered in neither list. Now guarded in both the workflow-step check and the local `ci_check.sh` list.
- `assert_mutants_command` captured `grep` output under `set -e`, so the exact regression it exists to catch (wrapper never reaching `cargo mutants`) died with no diagnostic. Now reports the log.
- Added assertions for the `--bins` baseline, the full `--no-shuffle`/`--shard`/`--sharding` triple, fractional timeouts, and the two rejection cases.
- `docs/RELEASING.md`'s `ci_check.sh` inventory refreshed — it now runs during a release.

### Deliberate scope note
Three findings required editing lines this PR did not originally touch: adjacent `README.md`/`CLAUDE.md` doc lines, the `Justfile` recipe comment that is the source of the wrong `--` convention, `scripts/test_ci_policy.py`, and `docs/RELEASING.md`. The widening was intentional — the `-- --diff` breakage is the same "documented invocation doesn't work" defect this PR exists to fix, and leaving it would have shipped a docs block teaching two incompatible conventions. Reviewer may narrow if preferred.

### Verification
`bash -n` on both scripts · `./scripts/test_run_mutants.sh` · `python3 -m unittest discover -s scripts -p 'test_*_policy.py'` (6 passed) · `./scripts/test_test_all.sh` · `just --list` · `just --dry-run mutants` for every documented form. No Rust sources changed, so the cargo build was not re-run.

pmatos:
## Follow-up: `--diff` branch coverage (7a1338d)

Self-review of the previous commit caught a false coverage claim I'd left in `scripts/test_run_mutants.sh`:

```
# The --diff branch needs a real git repository, so it is exercised by
# `just mutants --diff` rather than by this fixture.
```

Nothing in CI or `ci_check.sh` runs `just mutants --diff`, so the branch had **no** coverage. Asserting coverage that doesn't exist is the same defect class this PR is about, so I closed the gap rather than reword the comment.

The fixture now gets a real git repository (base ref distinct from HEAD) and asserts command construction for three paths:

| Invocation | Asserted |
|---|---|
| `--diff` | `--in-diff <DIFF>` appended, default timeout 180 |
| `--diff base` | explicit base ref consumed, not rejected as unknown argument |
| `--diff --timeout 45` | optional base ref does **not** swallow the following flag |

The mktemp path is normalised to `<DIFF>` so assertions pin the flag, not the random filename.

**Teeth check.** I confirmed the new assertion actually fails on the bug it targets: removing the optional-base guard (`"$1" != -*`) from `run-mutants.sh` makes `--diff --timeout 45` consume `--timeout` as the base ref, and the suite fails with `unknown argument: 45`, exit 2. The wrapper was restored from HEAD afterwards (verified clean).

Good news for the guard itself: it was already correct on this branch — the check found no bug, only the missing test.

### Verification (re-run, all forms this time)
`bash -n` · `./scripts/test_run_mutants.sh` · policy unittests (6 passed) · `./scripts/test_test_all.sh` · `yaml.safe_load` on `test.yml` (the workflow step is newly block-form, so it's worth parsing) · `just --dry-run mutants` for all six documented forms — bare, `--diff`, `--diff main`, `--shard 0/8`, `--timeout 45`, `-- --foo`. My earlier comment said "every documented form" while I had in fact only run a subset; that's now literally true.

Still no Rust sources touched, so the cargo build was not re-run.

pmatos:
## Addendum (54adb8a): passthrough assertion + fixture git isolation

Two last gaps in the file this PR adds, both closed:

- **`-- --foo` passthrough was unpinned.** This PR changed the documented spelling from `just mutants -- -- --foo` to `just mutants -- --foo`. I'd verified the `just` half with `--dry-run`, but nothing pinned the wrapper half. Traced it — `args+=("${extra[@]}")` in `run-mutants.sh` forwards correctly, so no bug — and added an assertion for `--timeout 45 -- --foo --bar` so the flag ordering (wrapper flags first, passthrough last) is now a contract.
- **Fixture git inherited ambient config.** The `--diff` fixture repo I added in 7a1338d ran `git init`/`commit` against the developer's real global config. A global `core.hooksPath`, `init.templateDir`, or `commit.gpgsign` could change the test's behaviour machine to machine. All fixture git commands now run with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null` and repo-scoped identity.

`scripts/test_run_mutants.sh` now covers the wrapper's full argument surface: default timeout, override, fractional, `--shard` triple, all three `--diff` paths, `--` passthrough, the `--bins` baseline, and three rejection cases.

### Status
Green locally: `bash -n`, `./scripts/test_run_mutants.sh`, policy unittests (6), `./scripts/test_test_all.sh`, workflow YAML parse, and `just --dry-run mutants` for all six documented forms. Working tree clean.

GitHub Actions had not yet registered runs for the three new commits when this run ended — worth a glance at the **Test** job in particular, since its mutation-wrapper step is new and now runs before `Install dependencies`. I validated that step two ways locally (the workflow YAML parses, and the exact command passes), but it has not executed in the hosted environment yet.

pmatos:
## `/simplify` pass — 1 fix applied, structural collision flagged

Ran a four-angle simplification pass (reuse / simplification / efficiency / altitude) over `origin/main...HEAD`.

### Applied

**`CLAUDE.md` / `README.md` — removed a literal that an already-merged CI gate forbids.** (a2990fc)

`scripts/test_mutants_invocation.sh`, merged to main in #764, asserts that `Justfile`, `CLAUDE.md` and `README.md` never contain the string `just mutants -- <two-dash>diff`. This branch's new passthrough prose embedded that exact literal in both files, so the doc-drift gate fails the moment the branch sees current main. Rephrased to describe the separator's effect rather than spell out the invalid invocation. Verified against main's assertions in both directions — the forbidden literals are gone and the required ones (`just mutants --diff`, `just mutants --shard 0/8`, `just mutants -- --foo`) are all still present. `test_run_mutants.sh`, the CI-policy unittests, `test_test_all.sh`, and main's four `assert_dry_run` cases all pass.

### Blocker for merge (not fixed here — out of scope for a simplify pass)

**This branch collides with #764.** `origin/main` is now `4f76668`. #764 landed overlapping scaffolding: a wrapper regression harness (`scripts/test_mutants_invocation.sh`), a workflow step, a `ci_check.sh` insertion at the same position 2 with the identical 2→3…6→7 renumbering, and the same bare-flag doc corrections. `git merge-tree` reports content conflicts in **`CLAUDE.md`, `Justfile`, `README.md`, `ci_check.sh`**.

The `--timeout` feature itself is the genuinely new and valuable part — #764 touches neither `scripts/run-mutants.sh` nor `.cargo/mutants.toml`.

The two harnesses assert at different layers, and both layers are worth keeping:
- main's asserts at `just --dry-run mutants` — the documented entrypoint #218 is actually about;
- this branch's asserts the resulting `cargo mutants` argv via a PATH-stubbed `cargo`, which `just --dry-run` cannot see.

**Recommendation:** fold this branch's argv assertions into main's existing `scripts/test_mutants_invocation.sh` rather than keeping two harnesses, two workflow steps and two competing "step 2"s in `ci_check.sh`. That resolution needs a rebase and a real restructuring decision, so I did not attempt it unattended.

### Reviewed and deliberately skipped

- **`--shard`/`--timeout` validation duplication** (`run-mutants.sh`). The new `--timeout` block is a line-for-line copy of the pre-existing `--shard` block. A validate-only helper taking `(flag, regex, description, example)` would unify them with byte-identical error strings and no nameref, but it edits a region conflict resolution will touch, and two instances is arguably not yet debt.
- **Log-path threading in `test_run_mutants.sh`** (eight write-once `*_log` vars; `assert_rejected` re-spells `run_wrapper`'s invocation). Real, but this file is the most likely to be merged away when the collision is resolved.
- **The `--baseline=skip` rationale duplicated across four files.** Flagged by two agents. Not consolidating: this PR exists *because* that rationale was missing and wrong, and `.cargo/mutants.toml` and the wrapper header specifically want it in situ.
- **`ci_check.sh` hardcoded step numbers.** Genuine — both this branch and #764 paid the same renumbering tax and collide because of it, and step 5 sits behind an `if [ -f ./build_tests.sh ]` guard so output can jump "4." to "6.". A counter would make future insertions free. Deferred: most of this branch's `ci_check.sh` delta is already on main.
- **`--diff base` test case** asserts a byte-identical string to the `--diff` case (both refs point at the same commit), so it only proves the positional isn't rejected as an unknown argument. Kept — weak coverage beats none, and strengthening it is a test-quality change, not a simplification.

**No efficiency findings** — the harness measures 0.12s / ~8MB, never invokes real cargo, and its fixture setup is correctly hoisted.

pmatos:
## Merge conflicts with `main` resolved

`main` picked up #764 (`fix(ci): document mutants wrapper flags without separator`), which independently corrected the same `just mutants` documentation this PR rewrites and added its own wrapper regression script. Resolved in merge commit `a41ba64`.

### Judgment calls

**Merge, not rebase.** `main` was one commit ahead, and the conflicting doc regions (`CLAUDE.md`, `README.md`, `Justfile`) are touched by three separate commits on this branch — a rebase would have replayed the same resolution 2–3 times against partially-built intermediate states. The repo squash-merges, so branch history does not reach `main` anyway and a merge commit costs reviewers nothing while avoiding a force-push on a published branch.

**Both regression scripts kept.** They cover different layers and neither subsumes the other:
- `scripts/test_mutants_invocation.sh` (from `main`) pins the `just` → wrapper hand-off via `just --dry-run`, plus the documented spellings in `Justfile`/`CLAUDE.md`/`README.md`.
- `scripts/test_run_mutants.sh` (from this PR) pins the `cargo mutants` argument vector the wrapper actually builds, including `--timeout` parsing and rejection paths.

Folding them into one script would be a refactor, not a conflict resolution, so both stay.

**Two CI steps at two distinct positions — deliberate, not an oversight.** In `.github/workflows/test.yml`, `Check mutation-test wrapper invocations` stays *after* `Install dependencies` because it shells out to `just`, which that step installs. `Check mutation-wrapper command construction` stays *before* it, since it only needs bash/git/mktemp and stubs `cargo`/`cargo-mutants` on `PATH` itself. Collapsing them into one location would pass locally (where `just` is already installed) and fail in Actions with `just: command not found`.

**`ci_check.sh` renumbered to 8 steps** (policy → invocation regression → command regression → fmt → build → test binaries → tests → `test_all.sh`), and `CLAUDE.md`'s enumeration of that script updated to match. #764 added its step without updating the `CLAUDE.md` list, and nothing in `scripts/test_ci_policy.py` checks the enumeration against the script, so that drift would otherwise have gone unflagged.

**Conflicting prose: this branch's version kept.** In `CLAUDE.md` and `README.md` the two sides say the same thing about bare-vs-separator flags; this branch's text is a strict superset (it also documents `--timeout`, `--in-place`/`--baseline=skip`, and why `.cargo/mutants.toml`'s `timeout_multiplier` does not apply to the wrapper). It names `.cargo/mutants.toml` explicitly, so dropping `main`'s shorter "Configuration lives in …" bullet loses no information.

**Not done, on purpose:** no `test_ci_policy.py` entry was added requiring `scripts/test_mutants_invocation.sh` as a gate. #764 shipped without one, and adding it is new scope rather than conflict resolution. Worth a follow-up if the policy test should cover both wrapper gates symmetrically.

### Verification

Full local gate green on the merged tree — all 8 `./ci_check.sh` steps pass, including both wrapper regression scripts and the complete test suite.

PR #770, opened 2026-08-24
test(aarch64): explain extended LDRS fixture exclusion

### LINKED ISSUE #513: Comment the Extended/Ldrs exclusion in the single_reg_instructions test helper
Deferred from #495: add a one-line comment to the `single_reg_instructions` test helper explaining why `AccessWidth::Extended` is excluded from the generated `Ldrs` instruction.

Original feedback from the automated `claude[bot]` code review on PR #495 (https://github.com/pmatos/s11/pull/495):

> `AccessWidth::Extended` is excluded from `Ldrs` in the test helper (matching the existing constraint that LDRSQ doesn't exist). The comment in `is_encodable_ldr_like` already documents this as constraint (d); the test helper silently enforces it via the `if width != AccessWidth::Extended` guard. This is correct but could benefit from a one-line comment in the helper for future readers.

**Context:** `src/ir/instructions.rs`, the `single_reg_instructions` test helper in the `tests` module — the `if width != AccessWidth::Extended` guard that omits the `Ldrs` arm (signed 64-bit load LDRSQ does not exist).

**Why deferred:** diminishing-returns — a cosmetic one-line comment in a test helper that the reviewer itself flagged as optional polish; the repo's `claude-review` job churns on cosmetic-only pushes, so it is not worth changing in this PR.

pmatos:
## Summary

- explain why the `single_reg_instructions` helper skips `Ldrs` for `AccessWidth::Extended`
- document the architectural absence of `LDRSX` without changing test behavior

## Testing

- `cargo test --lib ir::instructions::tests::ldrs_rejects_extended_width`
- `cargo fmt --all`
- `cargo test --all` (after generating the required test binaries)
- `./ci_check.sh`

Closes #513

### DIFF
--- a/src/ir/instructions.rs
+++ b/src/ir/instructions.rs
@@ -2610,6 +2610,7 @@ mod tests {
                 width,
             },
         ];
+        // No LDRSX exists, so signed loads cannot use the extended width.
         if width != AccessWidth::Extended {
             instructions.push(Instruction::Ldrs {
                 rt: Register::X0,


pmatos:
## Automated code-review pass (`/code-review --fix`)

Ran a review pass over this PR's diff. **No fixes were applied** — the diff is a single comment line and nothing in it is defective.

Verified as part of the pass:
- The comment is factually correct: `AccessWidth::Extended` is 8 bytes (`src/ir/types.rs:523`), and LDRSX does not exist in AArch64 (64-bit sign-extension is the identity, so LDR covers it).
- No doc drift — the wording matches the canonical rule in the `is_encodable_ldr_like` doc comment and the `Instruction::Ldrs` arm of `is_encodable_aarch64` (`src/ir/instructions.rs:1198`), which rejects Extended before delegating.
- The skip is self-checking: inverting the condition to `==` feeds an `Ldrs { width: Extended }` into `assert_single_reg_encodable`'s positive-assertion loop, which fails.
- The parallel invariant is already enforced for pairs — `is_encodable_pair` rejects `signed && width != PairAccessWidth::Word` (line 1501), gating LDPSW.
- `cargo fmt -- --check` is clean.

### One out-of-scope finding (not fixed)

`src/ir/instructions.rs:1447` — `is_encodable_ldr_like` contains a no-op `if width == AccessWidth::Extended { }` whose body is comment-only. It evaluates the condition, executes an empty block, and falls through to `true` exactly as if it were absent. The real Extended rejection lives ~250 lines earlier in the `Instruction::Ldrs` arm.

Deliberately left alone: it sits ~1170 lines outside this diff, in a different function, and the comment inside the block is the only place explaining why the shared helper permits Extended while the variant arm narrows it. Dedenting it out of the block would change what a reader understands it to scope over — that is an editorial decision on code this PR does not touch. Worth a follow-up issue, not a drive-by edit here.

Related observation for that follow-up: this invariant is now stated in seven places in `src/ir/instructions.rs` (the `is_encodable_ldr_like` doc comment, the `Ldrs` arm at 1198, the dead block at 1448, `ldrs_mnemonic`'s doc at 1878 and its `unreachable!` at 1888, `ldrs_rejects_extended_width` at 2865, and the new line 2613). Any consolidation would happen in the other six sites, all out of diff.

pmatos:
### Simplify pass — no changes applied

This automated simplify pass fired **after** the PR had already merged
(merged `2026-08-24T18:01:52Z`, merge commit `ec8f0f6`). Local branch HEAD
`a2a8a5c` is identical to the merged `headRefOid`, so there was nothing
unlanded to simplify onto.

The pass still ran against the merged diff. Reviewed for reuse,
simplification, efficiency, and altitude:

- **Reuse** — the architectural fact ("no LDRSX exists") is now stated in
  four places in `src/ir/instructions.rs`: the encodability rule (`:1197`),
  a helper doc comment (`:1879`), the `ldrs_rejects_extended_width` test
  (`:2864`), and the line this PR added (`:2613`). The first three all
  pre-date this PR, so the triplication is pre-existing and consolidating it
  would mean changes well outside the reviewed diff. The added line is the
  explicit deliverable of #513 and helps a reader sitting at the fixture, so
  it stays.
- **Simplification** — the diff adds no logic, only a comment. Nothing to
  simplify.
- **Efficiency** — no runtime or test behavior changed.
- **Altitude** — correct depth. The local fixture exclusion is explained at
  the fixture; the authoritative rule already carries its own rationale at
  `src/ir/instructions.rs:1197`.

**Result: no fixes to apply, nothing committed or pushed.** The working tree
is clean and matches the merged state.

Judgment call recorded per the operating contract: this run exits 0 rather
than flagging blocked. `gh pr list --state open` returns nothing for this
branch, which the run contract cites as a blocked condition — but here it
means the PR merged successfully, not that the simplify target was never
created. Exiting non-zero would mark a landed issue as failed. An operator
can override if the intended semantics were otherwise.

PR #771, opened 2026-08-24
test(parser): cover label before directive

### LINKED ISSUE #514: Add parse test for label-before-directive (_start: .text → Skip)
Deferred from #487: add an explicit regression test for the label→directive three-way interaction, e.g. `_start: .text` should parse as `Skip`.

Original feedback by the automated code review on PR #487 (https://github.com/pmatos/s11/pull/487):

> One optional addition: test label-before-directive (e.g. `_start: .text` -> Skip) to make that three-way interaction explicit.

**Context:** `src/parser/mod.rs` — the `strip_leading_labels` → `is_directive` ordering in `parse_line`. `_start: .text` already strips `_start` to leave `.text`, which `is_directive` then skips; a test would lock that in.

**Why deferred:** diminishing-returns — the behavior already works and adjacent label-stripping/directive paths are already covered, so this is an optional nicety rather than a coverage gap. The suggestion is explicitly optional on an already-approved, green PR, and a test-only push would trigger another review-bot churn round. The two evaluators split DEFER/REJECT.

pmatos:
Adds a focused parser regression proving that `_start: .text` is classified as `LineResult::Skip` after label stripping and directive recognition.

Closes #514

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -2636,6 +2636,14 @@ mod tests {
     }
 
     #[test]
+    fn parse_line_skips_inline_label_before_directive() {
+        assert!(matches!(
+            parse_line("_start: .text").unwrap(),
+            LineResult::Skip
+        ));
+    }
+
+    #[test]
     fn parse_line_accepts_multiple_inline_labels_before_instruction() {
         assert!(matches!(
             parse_line("outer: inner:").unwrap(),


pmatos:
Autonomous-run verification note: the generic `scripts/full_test.sh` named in the implementation-stage contract is not present in this repository, so I used the repository-authoritative `./ci_check.sh`, which passed in full. The required `cargo clippy --all -- -D warnings` also passed. An additional `--all-features --all-targets` clippy probe reports two pre-existing `clone_on_copy` warnings in untouched `src/x86_window_reassembly.rs` lines that are identical to `origin/main`; I left those unrelated baseline warnings out of this focused parser-test PR.

pmatos:
Automated `/code-review --fix` pass (target: this PR, base `origin/main`, effort level: medium as invoked; the review ran at xhigh).

**Result: no findings, nothing fixed, nothing pushed.** The diff is 8 lines — one new unit test in `src/parser/mod.rs`. The review confirmed the test is discriminating (mutating `strip_leading_labels` to identity, or checking `is_directive` before the strip, both make it fail), matches the `assert!(matches!(..))` convention of the adjacent inline-label tests, and passes alongside the other 17 `parse_line_*` tests. All 11 PR checks are green.

Two out-of-scope observations, deliberately **not** acted on:

1. `parser::x86::parse_x86_assembly_string` (`src/parser/x86.rs:805`) has no inline-label stripping, so `_start: .text` and `_start: mov eax, ebx` fail on the x86 path where AArch64 accepts them — and its doc comment's claim to mirror `parse_assembly_string` is therefore inaccurate. Pre-existing and unrelated to this diff; fixing it is a behaviour change that deserves its own issue.
2. The PR body says to merge with `gh pr merge --merge` "per CLAUDE.md". Neither the repo `CLAUDE.md` nor the global one states any merge-strategy rule, so that citation is unfounded — please use whatever this repo's actual convention is rather than treating that line as policy. Left the body unedited rather than rewriting it mid-review.

Issue #619, opened 2026-06-23
--auto driver: indirect control-flow soundness (jump tables, relocs, rodata pointers)


pmatos:
Part of #615. Implements **ADR-0009 Decision 5**. Builds on the direct-target-exclusion slice. **This is the soundness gate for running `--auto` on real binaries (e.g. `/bin/ls`).**

## Background

Direct branch targets are recoverable by linear scan. **Indirect** control flow
is not: jump tables from `switch` lowering, PLT stubs, computed gotos, and
function pointers have targets that never appear as a direct-branch operand. If
the driver rewrites a window whose interior is secretly an indirect-jump target,
the produced binary segfaults or misbehaves only on the inputs that take that
edge — the worst failure mode this tool can have.

The rule is **conservative refusal**, never guessing. Fail closed.

## Scope

Extend the branch-target set with:

- Any address named by a **relocation** (treat as a potential entry point).
- Any value stored in `.rodata` / `.data.rel.ro` that falls inside an executable
  section's address range (a candidate jump-table entry / code pointer).

Then refuse any candidate window whose interior contains such an address.

- This **over-approximates** — it rejects some genuinely-safe code near tables.
  That is the intended trade: soundness first.
- `log()` the count of windows refused on this basis so suppressed coverage is
  never silent (ADR-0009 Decision 5 & 9).

## Acceptance criteria

- Fixture with a `switch`-style jump table: a window straddling a table target is
  refused.
- Relocation-named and `.rodata` code-pointer addresses are included in the
  exclusion set.
- The count of windows refused for indirect-target reasons is logged.

## Soundness note

Unknown / ambiguous → **refuse**. Until this slice lands, `--auto` must not be
run on any binary that may contain indirect control flow (i.e. essentially all
real binaries).

## Code pointers

- ELF relocation sections / symbol access — `src/elf_patcher/`.
- Read-only data section ranges and executable section address ranges.
- The target set built in the direct-target slice.

pmatos:
Implementation design after re-deriving the missing workspace `PLAN.md` from ADR-0009:

- Keep direct-target behavior unchanged (split at instruction boundaries), then **refuse the entire planned candidate** when an indirect target lies strictly inside it. A target equal to the window start remains allowed. This follows the issue's refusal wording and gives an unambiguous refusal count; splitting indirect candidates was considered but rejected because it would weaken the explicit conservative gate.
[...]

PR #776, opened 2026-08-24
feat(cli): implement whole-binary auto loop driver

### LINKED ISSUE #620: --auto driver: loop driver (worklist, no-improvement cache, termination)
Part of #615. Implements **ADR-0009 Decisions 3, 8, 9** — the loop that ties find → pick → optimize → patch → repeat together.

## Background

With the CLI surface, window-finder, and direct-target exclusion in place, this
slice is the actual driver. It reuses the existing per-backend
`optimize_elf_binary_with_backend` for the inner search and the existing patcher
for replacement, operating on an **in-memory image** and writing the final
result to `-o`.

## Scope

- Drive the five-step loop over an in-memory image of the binary.
- **Monotone worklist**: each accepted rewrite strictly lowers the chosen
  `--cost-metric`, so progress is monotone; terminate when a full pass accepts
  zero rewrites.
- **No-improvement cache** keyed by a hash of the window's instruction bytes:
  skip a window whose bytes are unchanged since a prior miss (avoids re-proving
  across passes and overlaps).
- **Prioritized selection** (ADR-0009 Decision 9): prefer longer admissible runs
  / windows with apparent redundancy; honour a global budget (time and/or window
  count) and `log()` what was skipped when the budget bounds coverage.
- Per-window search time bounded by the existing `--timeout`.
- Write the patched image to `-o` (falling back to the derived path if `-o`
  omitted, per the CLI slice).

## Dependencies

- **Requires**: CLI surface slice, window-finder slice, direct-target slice.
- **Strongly recommended before running on real binaries**: the indirect-jump /
  jump-table soundness slice (without it, refuse to run on sections containing
  indirect control flow, or gate behind a flag).
- **Enhanced by**: downstream register live-out slice, x86 latency cost-model
  slice (both optional for correctness).

## Acceptance criteria

- End-to-end on a crafted fixture ELF with a couple of optimizable supported
  windows: the output binary differs only within those windows, is NOP-padded to
  original length, and **all downstream bytes are unchanged**.
- Re-running on the output reaches a fixpoint (zero further rewrites).
- The no-improvement cache demonstrably prevents re-proving an unchanged window.
- A bounded run logs the number of windows skipped due to budget.

## Code pointers

- `optimize_elf_binary` / `optimize_elf_binary_with_backend` — `src/main.rs`.
- The patcher entry that writes a window and NOP-pads — `src/elf_patcher/`.

pmatos:
## Summary

- add an ISA-agnostic monotone worklist with exact-byte miss caching, prioritized selection, strict cost-decrease enforcement, and a global `--max-windows` budget
- patch an in-memory ELF image across passes and write the final result once, while using ADR-0009 per-window indirect-target exclusions and reporting refused coverage
- fail closed on relocatable (`ET_REL`) objects whose executable sections cannot be identified safely by the v1 address-only worklist
- cover multi-window rewriting, NOP padding, downstream-byte stability, no-improvement cache hits, budget reporting, byte-identical fixpoint reruns, and pre-write `ET_REL` refusal

## Verification

- `cargo test --all`
- `./ci_check.sh`

Closes #620


pmatos:
## Automated code-review pass (`/code-review --fix`, effort: xhigh)

Pushed as 9107d73. CI gate (`./ci_check.sh`) green before push.

### Applied

| Area | Fix |
|---|---|
| `src/main.rs` | Run summary no longer claims "reached a fixpoint" when whole executable sections were refused. Refused candidate windows are counted (`refused_windows`) and reported as incomplete coverage. |
| `src/main.rs` | The per-section indirect-control-flow refusal message is deduplicated (`reported_refusals`). Discovery reruns after every accepted rewrite, so a binary with one refused section and 20 rewrites previously emitted 21 identical multi-line messages. |
| `src/main.rs` | Indirect-flow detection folded into the existing phase-1 `insn_detail` loop and carried on `SectionCandidateWindows`; `indirect_control_flow_addresses` deleted. It had been re-reading and re-disassembling every executable section on every pass, duplicating work the same call had just done. |
| `src/elf_patcher/mod.rs` | `apply_patch` now bounds-checks against the file length instead of panicking. A section header naming a range past EOF passed `validate_address_window` and then hit an index-out-of-bounds abort from a `pub fn` returning `Result`. |
| `src/elf_patcher/mod.rs` | Follow-on (8afa4d4): the `section.file_offset + offset_in_section` u64 add above that check was itself unguarded. `validate_address_window` constrains only a section's `virtual_addr` and `size` — it never touches `file_offset` — so `sh_offset` reaches `apply_patch` attacker-controlled. A wrapping add in a release build would have produced a small offset that *passes* the new file-length check and writes the patch to the wrong place. The add and the `usize` conversion are now both checked. |

### Deferred — needs author decision

**1. `--auto` is currently inert on any binary containing `ret`.** `capstone_detail_is_indirect_control_flow` (`src/main.rs:1330`) includes `CS_GRP_RET` in its `transfers_control` set, and a `ret` carries no immediate operand, so `capstone_detail_direct_branch_targets` returns empty and the instruction is classified as unresolvable indirect flow. Every section containing a return is therefore refused.

Reproduced directly against `9107d73` on both arches:

```
$ s11 opt --auto binaries/x86_64/simple_opt --output /tmp/o1
Auto mode refused executable section '.init' ... skipped 3 candidate window(s) ...
Auto mode refused executable section '.text' ... skipped 9 candidate window(s) ...
Auto mode refused executable section '.fini' ... skipped 1 candidate window(s) ...
Auto summary: 0 searched, 0 cache hits, 0 rewrites accepted.
Auto coverage is incomplete: refused 13 candidate window(s) ...

$ s11 opt --auto binaries/simple_opt --output /tmp/o2
... .init (2), .plt (7), .text (21), .fini (1) ...
Auto summary: 0 searched, 0 cache hits, 0 rewrites accepted.
Auto coverage is incomplete: refused 31 candidate window(s) ...
```

Both exit 0, and `cmp` reports the output byte-identical to the input. After this PR they at least say so out loud (the incomplete-coverage line above is the fix), but the feature itself does nothing on a normal binary.

Not fixed here: loosening the gate is a soundness call tied to #619, not a mechanical cleanup. The argument for excluding `ret` is that a return target is the instruction after a `call`, and calls are already `WindowRole::Excluded`, so return sites always begin a fresh window. The argument against is that retpolines, `longjmp` thunks, and `ret`-based tail-call trampolines use `ret` as a genuine computed jump.

**2. One bad window aborts the whole run and discards every accepted rewrite.** `drive_auto_optimization(...)?` sits above `image.write_to(output_path)?` in `run_auto_optimization_with_backend`, so any window-level error propagates out before the image is written. Three sources reach it: `reassemble_optimized_x86_window` refusing a byte-longer encoding (nothing in the search bounds encoded byte length — under the default `--cost-metric instruction-count` a 1-instruction x86 replacement carrying an imm32 can exceed a 2-instruction window), `apply_patch`'s "New code is larger than window size", and the monotonicity gate. With `--max-windows 100`, window 97 tripping any of these throws away 96 verified rewrites and writes no output file. These should cache the window as a miss and continue; changing that is a failure-contract decision.

**3. The strict-cost-decrease gate is fatal when a backend's acceptance criterion differs from `--cost-metric`.** `src/auto_driver.rs:111` returns `Err` (fatal, per #2) when `optimized_cost >= original_cost`. The LLM backend gates purely on `candidate.len() >= target.len()` (`src/search/llm/outcome.rs:59`) and never consults the cost metric, so `--auto --algorithm llm --cost-metric latency` can hand back a shorter-but-costlier sequence (two MOVs, latency 2 → one MADD, latency 3) and kill the run. A window the search over-claimed is arguably a miss, not an invariant violation.

### Deferred — scope or trade-off

- **No-improvement cache key** (`src/auto_driver.rs:87`) is the window's own bytes, but a search result also depends on downstream liveness. A later rewrite that shrinks an earlier window's live-out set leaves that window's bytes unchanged, so it hits the stale cached miss and is never re-searched; symmetrically, two byte-identical windows with different downstream contexts share one entry.
- **No global wall-clock budget.** `timeout` has no clap default, so `s11 opt --auto prog` runs up to `--max-windows` searches with no per-search deadline. ADR-0009 Decision 9 says the driver honours "time and/or window count" — only the count exists. Combined with the write-at-the-end structure, Ctrl-C after 50 accepted rewrites yields no output.
- **Per-window reporting noise.** `optimize_elf_window_with_backend` is reused verbatim in the loop, so the full single-window report (header, disassembly, IR dump, search statistics) is reprinted per window and buries the four-line summary. It also prints the input path while operating on the mutated in-memory image.
- **`discover_windows` re-parses the ELF per candidate.** `get_instructions_in_window` re-runs `ElfBytes::minimal_parse` plus a section-header scan and re-decodes the window, even though phase 1 just decoded every instruction in the section — O(W×R) full-ELF parses. `instruction_count` and `instruction_bytes` are derivable from the phase-1 decode.
- **NOP-padding remnants burn budget.** Each accepted rewrite leaves a 1-instruction remnant that is re-discovered, searched, and charged against `--max-windows`. On this PR's own fixture, `--auto auto_two_dup_mov --max-windows 10` reports "4 searched, 2 rewrites accepted". Enumerative and symbolic search provably bail below 2 instructions, but MCMC's `min_length = 1` means a blanket filter would lose real stochastic/code-size rewrites — hence not fixed.
- **Test coverage gap.** The only end-to-end `--auto` fixture (`tests/x86_asm/auto_two_dup_mov.s`) is a `-nostdlib` `_start` with no `ret` — exactly the shape that dodges the refusal gate in finding 1. An assertion that `--auto` on any `binaries/x86_64/*_opt` reports a non-zero searched count would have caught it. The test also early-returns green when `binaries/` is unpopulated.
- **Backend dispatch duplicated three ways.** The `match patcher.arch()` block now appears in `find_candidate_windows` (test-only, `allow(dead_code)`), `optimize_elf_binary`, and `run_auto_optimization`.

pmatos:
### Auto mode refuses every section of a real binary (`ret` classified as indirect control flow)

Found while running the `/simplify` pass on this PR. Not a simplification finding — flagging it here rather than fixing it, since it is a behavioural/design question and out of scope for a quality pass.

`capstone_detail_is_indirect_control_flow` (`src/main.rs:1329`) returns `true` when the instruction is in `CS_GRP_JUMP | CS_GRP_CALL | CS_GRP_RET | CS_GRP_IRET` **and** `capstone_detail_direct_branch_targets` yields nothing. A plain `ret` carries no immediate operand, so it always lands in that bucket — and `discover_windows` refuses the *entire enclosing section* on a single hit.

Empirically, against the existing `binaries/x86_64/simple_opt` fixture:

```
$ s11 opt --auto simple_opt --max-windows 0 -o out.elf
Auto mode refused executable section '.init' because indirect control flow at 0x1014, 0x101a ...; skipped 3 candidate window(s)
Auto mode refused executable section '.text' because indirect control flow at 0x1025, 0x104f, 0x107f, 0x1088, 0x10c0, 0x10c8, 0x10f2, 0x1105, 0x1110 ...; skipped 9 candidate window(s)
Auto mode refused executable section '.fini' because indirect control flow at 0x1138 ...; skipped 1 candidate window(s)
Auto summary: 0 searched, 0 cache hits, 0 rewrites accepted.
Auto coverage is incomplete: refused 13 candidate window(s) ...
```

Seven of those twelve addresses are a bare `ret`:

```
1014: ff d0            call *%rax
101a: c3               ret
1025: c3               ret
104f: ff 15 6b 2f 00 00 call *0x2f6b(%rip)
107f: ff e0            jmp  *%rax
1088: c3               ret
10c0: ff e0            jmp  *%rax
10c8: c3               ret
10f2: ff 15 e8 2e 00 00 call *0x2ee8(%rip)
1105: c3               ret
1110: c3               ret
1138: c3               ret
```

Consequence: on any ordinary ELF, `--auto` performs zero searches and zero rewrites, and the `"reached a fixpoint"` line (gated on `fixpoint_reached && refused_windows == 0`) can never print. The new integration test does not catch this because `tests/x86_asm/auto_two_dup_mov.s` is a `_start` with no `ret`, `jmp`, or `call` at all.

The soundness property the gate is protecting is "no control transfer may land at an address inside a candidate window that we did not account for". `ret` and indirect `call` transfer control *out of* the section (to a return address / callee); they do not create an unknown in-section entry point. An indirect `jmp` (jump table, computed goto, PLT thunk) genuinely can. Narrowing the predicate to indirect **jumps** only — dropping `CS_GRP_RET`, `CS_GRP_IRET`, and indirect `CS_GRP_CALL` — would preserve the property while letting auto mode actually run. Alternatively, refuse only the affected windows rather than the whole section.

Either way this seems worth resolving before `--auto` is useful outside the synthetic fixture. Deliberately not changed in the simplify commit.

pmatos:
### `/simplify` pass — findings deliberately skipped

Applied fixes are in 2f02c29 and a3e0856. Four review agents (reuse, simplification, efficiency, altitude) raised the following, which I did **not** act on. Recording them so a future reviewer doesn't re-derive them.

**Would change behaviour**

- **Remove `ElfWindowOptimization::NoImprovement`'s byte payload entirely.** Tempting — the field is caller-specific — but single-window `s11 opt --start-addr/--end-addr` writes an output file *even when search misses*, by reassembling the original into a patched copy. Dropping the payload stops that file being emitted. Renamed `single_window_bytes` → `reassembled` instead, and kept the policy in the single-window driver where it belongs.
- **Move the `optimized_cost < original_cost` gate into `optimize_elf_window_with_backend`** (altitude #4). The two drivers genuinely disagree today: single-window mode patches an equal-cost replacement, auto mode aborts the run. Reconciling them is a design decision about what `--auto` should do on an equal-cost rewrite, not a simplification.

**Correct but outside a quality pass's blast radius**

- **Thread instruction count / redundancy through `plan_candidate_windows`** so `discover_windows` stops re-reading and re-disassembling every candidate window each pass (reuse #3, efficiency #2, altitude #3). Real waste — N full ELF re-parses plus a second decode per pass — but the fix reaches into `src/candidate_windows.rs` and its unit tests, well outside this diff. Worth its own issue; the discovery-side cost is negligible next to SMT search today.
- **Cache per-section classification across passes** (efficiency #1). Same shape: `find_candidate_windows_with_backend` re-runs the Capstone→text→parser classification over the whole binary on every pass. Only matters once #619 stops sections being refused wholesale.
- **Move `refused_windows` into `AutoRunSummary`** (altitude #2, #6). Honesty about incomplete coverage is currently enforced at one call site rather than by the type. A genuine improvement, but it widens the `AutoOptimizationAdapter` trait's public seam (`discover_windows` would return windows + refusals) — that's an API redesign, not a cleanup.
- **Unify the three `DetectedArch` → backend dispatch ladders** (reuse #7, altitude #5). Each arm instantiates a different generic `B`, so the only real dedup needs a `macro_rules!` or a continuation trait. Both agents hedged on whether that beats a third `match`; I agree it's a judgement call, not a clear win.

**Judged not a finding**

- **`#[derive(Clone)]` on `ElfPatcher` copies the whole ELF image.** Flagged as an efficiency issue. It's one memcpy per auto run and one per single-window invocation — neither in a loop. `create_patched_copy` already cloned the same buffer before this PR.
- **Fully-qualified `std::collections::HashSet` in the new code.** Matches the existing idiom in `main.rs` (see the pre-existing `branch_targets` at the same call depth), so importing it would make the new code the odd one out.

Separately, see the earlier comment on `capstone_detail_transfers_control` classifying `ret` as indirect control flow — that one is a behavioural bug, not a simplification, and is still open.

pmatos:
## Conflict resolution against `main` (autonomous run)

`origin/main` was merged into this branch (commit `4ef77f2`), not rebased. Rebasing would have meant re-authoring three intermediate commits (`0b08d7c`, `9107d73`, `2f02c29`) so they each contain a feature the tip deletes — history documenting a decision reversed inside the same PR reads worse than one merge commit. `main` squash-merges, so the merge commit costs nothing downstream.

### What conflicted

`main` landed two auto-driver PRs that touch the same discovery path this branch reworks:

- **#775** — ADR-0009 Decision 5 indirect-target refusal (`ElfPatcher::indirect_control_flow_targets`, `refuse_windows_with_interior_targets`, `SectionCandidateWindows::indirect_target_refusals`). This **closed issue #619**.
- **#777** — incomplete AArch64 section tails, which changed the decoded-section shape to `(section, Option<Instructions>)`.

### Judgment call: the branch's section-level indirect gate was removed

This branch carried its own stopgap: any executable section containing an indirect control-flow *instruction* had **all** its candidate windows refused, with the comment "until issue #619 is implemented". #619 is now closed and ADR-0009 Decision 5 is implemented on `main`, so that precondition is satisfied — the gate was dropped in favour of `main`'s precise per-window refusal.

**The delta a reviewer should weigh:** this makes auto mode *less* conservative than this branch was. `main` refuses a window only when a *known* indirect target falls in its interior, where "known" is exactly ADR Decision 5's two sources — relocation-named addresses, and code pointers stored in `.rodata`/`.data.rel.ro`. An indirect jump whose target is computed purely at runtime is no longer refused. That is `main`'s shipped position and the ADR's stated over-approximation trade, not a new decision made here — but it is the point to override if the project wants the union instead.

**Alternatives considered and rejected:**

- *Keep both gates as a union.* Refuses strictly more than ADR Decision 5 requires and contradicts a closed #619; it would also permanently suppress whole sections (e.g. any `.text` with a single `ret`), which is the coverage cost Decision 5 was written to avoid.
- *Keep the branch's gate and drop `main`'s.* Discards the more precise, already-reviewed, already-merged mechanism.

### What was preserved from this branch

The branch's honesty requirement — "never claim a fixpoint over a binary whose coverage was incomplete" — is kept, now fed by `main`'s refusal count:

- `ElfAutoOptimizationAdapter::refused_windows` sums `section.indirect_target_refusals` per discovery pass.
- The summary prints `Auto coverage is incomplete: refused N candidate window(s) whose interior contained an indirect target.` when `N > 0`.
- A fixpoint reached with refusals outstanding is labelled `reached a fixpoint over admitted windows`; only a run with zero refusals prints the unqualified line.

The branch's `auto_adapter_refuses_section_containing_indirect_control_flow` test was retargeted to `auto_adapter_accounts_for_indirect_target_refusals`: same intent (refused coverage is accounted for, not silently dropped), now over `main`'s switch-table `.rodata` fixture. All three of `main`'s new discovery tests were kept as-is.

### Follow-on cleanups

- `SectionCandidateWindows::section` lost its only non-test reader when the removed refusal message went away; annotated `#[cfg_attr(not(test), allow(dead_code))]` with the reason rather than deleting a field the discovery tests pin.
- Two stale "until issue #619 lands" prose references updated (`README.md`, `tests/x86_asm/auto_two_dup_mov.s`). The three `#619` comments in `src/main.rs` are `main`'s own and were left untouched.

### Verification

- `./ci_check.sh` — green (policy checks, mutation-wrapper regressions, `cargo fmt --check`, build, test binaries, unit + integration, full suite).
- `cargo clippy --all-targets` — clean apart from two `clone_on_copy` warnings in `src/x86_window_reassembly.rs`, pre-existing on `main` (#758) and untouched here.
- `test_auto_x86_64_rewrites_two_windows_then_reaches_fixpoint` passes; the fixture produces zero refusals, so it still prints `2 rewrites accepted` and the unqualified `reached a fixpoint`.
- Manual `--auto` run on the fixture confirms `Auto summary: 4 searched, 0 cache hits, 2 rewrites accepted.` followed by the fixpoint line.

One behavioural note for reviewers: `main`'s `indirect_target_refusal_log` prints once per discovery pass, and auto re-discovers after every accepted rewrite, so an auto run now emits N+1 copies of that line.

pmatos:
## Automated code-review pass (`/code-review --fix`, high effort)

Five findings; three fixed in d76c7ab, two deliberately skipped.

### Fixed

- **One window's patch error aborted the entire `--auto` run** (`src/main.rs`). A cheaper-but-longer replacement — e.g. x86-32 `inc eax; inc eax` → `add eax, 2` under the default `instruction-count` metric — made `apply_patch` reject the rewrite; the error propagated out of `drive_auto_optimization` before `image.write_to`, so the run exited nonzero with no output file and every rewrite accepted so far discarded. `optimize_window` now checks the replacement against the window's own extent (`end - start`, the same number `apply_patch` validates), prints the refusal, and degrades it to `NoImprovement` so the byte cache records it and the loop still reaches a fixpoint.
- **Auto mode reassembled search misses it then discarded** (`src/main.rs`). AArch64's `assemble_window` re-encodes the original IR on every miss. `optimize_elf_window_with_backend` now takes `reassemble_on_miss`; single-window mode passes `true` (unchanged), auto passes `false`. This removes both the wasted work and an encoder-failure surface that could abort a whole-binary run over a window nothing was going to patch.
- **Budget comparison** in `src/auto_driver.rs` hardened from `==` to `>=` (only correct by construction today).

Plus: window failures now carry their address range in the error, and README documents the new per-window refusal.

### Skipped, with reasons

- **Length-1 windows spend `--max-windows` budget and inflate the skip count** (`src/candidate_windows.rs:108`). Real — the fixture shows `4 searched` for 2 useful windows, and `BudgetExhausted { skipped }` counts runs that can never improve. Not fixed here because window discovery policy is owned by ADR-0009 Decision 9 and the change would also alter non-auto behavior. Worth a follow-up issue if the skip-count accuracy matters.
- **`discover_windows` re-disassembles every candidate window** after the full-section pass, once per rediscovery pass (`src/main.rs:1461`). Pure redundant work, but removing it means threading decoded instructions out of `find_candidate_windows_with_backend` — a wider refactor than a review pass should land.

### Checked and cleared

AArch64 terminator re-encoding is correct across a shrinking prefix (`pc_relative_offset` recomputes the displacement from `base_address + idx*4` against the absolute `LabelId`), and `ElfPatcher` caches no parsed-ELF state, so cross-pass in-memory patching invalidates nothing.

### Verification

Full `./ci_check.sh` gate passes on the pushed tree: `cargo fmt --check`, build, and 1702 + 106 + 68 tests, plus the end-to-end AArch64 suite. Two new regression tests cover the fixes (`oversized_replacement_does_not_fit_its_window`, `auto_mode_skips_reassembling_a_search_miss`).

pmatos:
Two follow-up notes a reviewer may reasonably raise, considered and left alone:

- The new "does not fit" refusal is a bare `println!` and does not land in `AutoRunSummary` the way indirect-target refusals do (`9107d73`). The README wording deliberately claims no count, so nothing is misreported — but the asymmetry is real and would be a small, separable follow-up.
- `usize::try_from(candidate.window.end - candidate.window.start)` does the subtraction before the conversion. Discovery never builds `end < start`, so it cannot underflow in practice; left as-is rather than spending a fix cycle on an unreachable path.

Confirmed while verifying: `apply_patch` rejects on `(window.end - window.start) as usize` (src/elf_patcher/mod.rs:397), which is exactly the value the new pre-check gates on — the guard and the rejection agree on one number.

pmatos:
## /simplify pass

Ran a four-angle simplification review (reuse, simplification, efficiency, altitude) over `origin/main...HEAD` and applied the findings that survived triage. Quality only — no correctness hunting, and no behaviour change beyond output tidying. Full gate (`./ci_check.sh`) green, and `test_auto_x86_64_rewrites_two_windows_then_reaches_fixpoint` verified to actually run rather than skip on its fixture.

### Applied (75cc960)

| Change | Why |
|---|---|
| `find_candidate_windows` → `#[cfg(test)]` | Production reaches discovery via `find_candidate_windows_with_backend`; the compiler can now enforce what `allow(dead_code)` was asserting by hand. Flagged independently by 3 of 4 agents. |
| `Algorithm:` banner printed once per run | `run_auto_optimization_with_backend` already emits it, so auto mode reprinted it for every searched window (up to `--max-windows`). |
| Gate `indirect_target_refusal_log` on `refused > 0` | Discovery runs once per pass, so a clean binary logged `refused 0 window(s)` on every pass. |
| Extract `total_indirect_target_refusals` | Discovery and the auto adapter folded the same per-section counts independently — two derivations of one number that can drift. |
| Extract `auto_run_summary_log` + unit test | Matches the file's existing `indirect_target_refusal_log` / `incomplete_executable_section_tail_log` pattern. Three of these strings are the contract asserted by the integration test; the unit test now pins the wording so a reword fails locally first. |
| `optimize_elf_binary_with_backend` takes `&B` | Both sibling `*_with_backend` helpers already do. |
| `run_auto_optimization` takes `ElfPatcher` by value | The caller has no further use for it; the previous `patcher.clone()` kept a second whole-binary image resident for the whole run. |

### Considered and deliberately not applied

**Contradict decisions already recorded in this PR's comments** (from the review pass in `d76c7ab`), so re-litigating them would be net-negative:
- Replacing the `window_len` `usize::try_from` with `instruction_bytes.len()` — the comment deliberately gates on the window's own extent, the same number `apply_patch` validates against, rather than on the byte snapshot discovery took.
- Moving `fits_window`'s predicate into `ElfPatcher`.

**Design changes, not simplifications:** adding a `WindowSearchResult::Refused` variant plus a `refusals` counter (today an unencodable-in-place improvement is reported as `NoImprovement` and logged); threading `refused_windows` through the `AutoOptimizationAdapter` seam.

**Too large / outside this diff:** threading per-window `instruction_count`/`redundancy_score` out of `SectionCandidateWindows` to remove discovery's second disassembly pass (`WindowInstruction` carries no bytes, so `redundancy_score` can't come along for free); the duplicated section-suffix copies in `src/validation/downstream.rs`.

**Dropped on cost/benefit after a second look**, both of which were on the apply list first:
- Hoisting `indirect_control_flow_targets()` out of per-pass discovery. Sound in principle (patches only touch `SHF_EXECINSTR` bytes, while the scan reads relocations and `.rodata`/`.data.rel.ro`), but the loop is bounded by `accepted_rewrites` ≤ `max_windows` — ~100 ELF metadata parses against ~100 multi-second Z3 searches. Noise, not asymptotic pain.
- Rewriting `apply_patch`'s `if new_code.len() < window_size` as `if patch_end < window_file_range.end`. `window_size` survives either way (the guard above still uses it), so it removes a *use*, not a duplication — and it reads worse in offset space. It was also the only proposed change touching patching arithmetic.

**False positives**, each independently confirmed by two agents: `X86OptimizationBackend::sequence_cost`'s `match self.arch` (matches the house pattern at `disassembler()` and routes through `isa::CostModel` deliberately so the monotonicity gate can't drift from the search), and collapsing `ElfWindowOptimization` into `WindowSearchResult` (the `NoImprovement { reassembled }` payload is binary-only and does not belong in the lib type).

One nit left alone: dropping the unused `Clone`/`PartialEq`/`Eq` derives from `WindowSearchResult` and `AutoRunSummary` — Rust's API guidelines favour eager derives on public types.

pmatos:
Resolved the older P2 review about address-ambiguous relocatable ELFs in `2c988f0`.

The v1 auto worklist carries `AddressWindow` values without section identity, and multiple executable sections in an `ET_REL` object can share `sh_addr == 0`. Auto mode now retains the parsed ELF type and rejects `ET_REL` before candidate discovery or output writing. `ET_EXEC` and `ET_DYN` behavior is unchanged, as is existing single-window relocation analysis.

A focused regression went red on the old behavior (it wrote an output and claimed a fixpoint), then green with the guard. README and ADR-0009 document the boundary. Alternatives considered were threading section identity/file offsets through discovery and patching (larger redesign outside this loop slice) or rejecting `ET_REL` in `ElfPatcher::new` globally (would unnecessarily change non-auto consumers).

Verification after the fix: `cargo fmt --all`, `cargo clippy --all -- -D warnings`, `cargo test --all`, and `./ci_check.sh` all pass.

pmatos:
Disposition of the final automated-review observations:

1. **Fatal errors remain transactional.** Expected per-window non-results (including an optimized encoding that does not fit) are downgraded to `NoImprovement` and the run continues. An unexpected discovery/decoder/assembler/patch error still aborts without writing a partially optimized image. Writing an output while returning an error would make it ambiguous whether that image is a supported final result; changing to best-effort persistence is a separate failure-contract decision, not required by #620.

2. **The context-free exact-byte cache is intentional v1 behavior.** #620 and ADR-0009 Decision 8 explicitly require the no-improvement cache to be keyed by the window's instruction bytes, including across overlaps and addresses. Different downstream liveness can make this completeness-conservative (a valid optimization may be missed), but it cannot admit an unsound rewrite. The PR already exposes cache-hit accounting; expanding the key to liveness/window identity would be a different design decision.

3. **Identical cache keys cannot have different window lengths in this implementation.** `instruction_bytes` is the entire byte range returned by `get_instructions_in_window`, not only decoded non-padding instructions. Equal `Vec<u8>` keys necessarily have equal lengths, so a wider window cannot be suppressed by a narrower window's oversized-fit miss under the same key.

The duplicate disassembly and near-duplicate result enums are acknowledged non-blocking cleanup/scalability opportunities. No further change is needed for #620.

pmatos:
## Automated code-review pass (`/code-review --fix`, xhigh)

Reviewed the full `origin/main...HEAD` diff for this PR. Two findings were fixed and pushed as `7f9eba4`; the rest were left for a human call and are recorded here so they are not lost.

### Fixed

1. **A single window's search/reassembly error aborted the whole `--auto` run** (`src/main.rs`, `ElfAutoOptimizationAdapter::optimize_window`). Nothing reaches disk until the driver's final `write_to`, so propagating the error discarded every rewrite already accepted in memory and wrote no output binary at all. x86 window reassembly is a live source of these: it *fails* rather than returning oversized bytes when an optimized prefix would displace a pinned `Jcc`, so the `fits_window` gate that exists to prevent this outcome never sees that case. Now refused per-window, counted as `refused_rewrites`, reported in the run summary, and it qualifies the fixpoint the same way a refused window does. New regression test `auto_adapter_contains_a_window_search_failure_instead_of_aborting_the_run`.
2. **Stale `#[cfg_attr(not(test), allow(dead_code))]`** on `find_candidate_windows_with_detail_provider`, which this PR puts on the production discovery path. Removed.

`./ci_check.sh` passes end to end (fmt, build, clippy, `cargo test --bin s11` 110 pass, `cargo test --lib` 1703 pass, full suite).

### Deliberately not fixed — for a human call

- **Consequence of fix 1:** a *systemic* failure now exits 0 with a "fixpoint" summary instead of erroring. E.g. `--auto --algorithm llm` with the Codex CLI unavailable refuses every window and reports success. Worth a follow-up that distinguishes per-window from systemic failure (abort after N consecutive failures, or fail when `refused_rewrites > 0 && accepted_rewrites == 0`).
- **`auto_driver.rs:111`** — a non-strictly-decreasing candidate is a run-ending `Err`, losing all in-memory rewrites. `non_decreasing_candidate_is_rejected_before_image_mutation` pins that behaviour deliberately, so changing it to a per-window refusal is a design decision, not a review fix.
- **`auto_driver.rs:85`** — the no-improvement cache is keyed on window bytes alone and is never invalidated when a *neighbouring* rewrite changes a cached window's downstream liveness context. The loop can therefore declare a fixpoint over an image that still has an available rewrite. The same key also conflates "no cheaper equivalent exists" with "we declined to apply the one we found". Cache-key redesign, out of scope for a review pass.
- **`main.rs:1605`** — the `ET_REL` rejection sits on the `--auto` entry point, not on the address-window machinery it protects. The same virtual-address ambiguity is still reachable via single-window `s11 opt foo.o --start-addr ...`: `validate_address_window` returns the *first* section containing the address. Moving the guard into `ElfPatcher` would close it for both paths.
- **`auto_driver.rs:86`** — `cache_hits` counts hits per window per pass, so it grows with pass count and can exceed the candidate-window count ("62 cache hits" on a binary with 10 windows). Reporting nit.
- **`main.rs:1514`** — `discover_windows` re-reads and re-disassembles every candidate window phase 1 already decoded, and each `get_instructions_in_window` re-parses the whole ELF. Roughly O(windows x rewrites) full ELF parses on a real binary. Cheapest fixes: memoize the section table on `ElfPatcher`, and carry per-window instruction counts out of the phase-1 decode.
- **`main.rs:1757`** — single-window verbose output is emitted unconditionally per window in auto mode, ignoring `options.verbose`; discovery log lines also reprint every pass. Genuine altitude problem, but the integration test greps that stdout, so it is not safe to change in a review pass.
- **`main.rs:1755`** — caller mode is encoded twice (a `reassemble_on_miss: bool` parameter *and* an `Option` payload that is always `None` for one caller).
- **`main.rs:1142`** — `find_candidate_windows` is `#[cfg(test)]` rather than deleted, leaving a test-only arch dispatcher that duplicates production's own `match`; they can drift silently.
- **Test coverage** — the only end-to-end auto test uses a terminator-free x86-64 fixture. The pinned-`Jcc` reassembly path (the source of fix 1), AArch64 `--auto`, and every refusal path are uncovered end to end.
- **`README.md:102`** — `--max-windows` bounds search count but not wall clock, so `--timeout` multiplies by the budget (default worst case 100 x timeout). Accurate but easy to misread as a run budget; ADR-0009 Decision 8 scopes it per-window on purpose.

pmatos:
Current-main integration note: origin/main moved the late hard-link/TOCTOU output guard into ElfPatcher while this PR moved all final image writes behind ElfPatcher::write_to. The merge preserves both intents by storing the pinned input Handle behind Arc so cloned in-memory patchers share identity safely, and by applying the guarded single-handle truncate/write in write_to. This protects both single-window and --auto output paths. Targeted alias-refusal and auto-driver tests passed, followed by clippy, cargo test --all, and ./ci_check.sh.

PR #779, opened 2026-08-24
fix(x86): reject REX-only byte registers in 32-bit mode

### LINKED ISSUE #684: Tighten X86Register::is_available_in to reject REX-only low-byte registers (SPL/BPL/SIL/DIL) in x86-32
Deferred from #681: `X86Register::is_available_in(Mode32)` returns `true` for the REX-only low-byte registers SPL/BPL/SIL/DIL (LowByte views of RSP/RBP/RSI/RDI, indices 4–7), even though they are unencodable in 32-bit mode.

Original feedback by @chatgpt-codex-connector (P2) and the claude-review bot on PR #681 (https://github.com/pmatos/s11/pull/681):

> When a configured register pool contains SPL, BPL, SIL, or DIL, this predicate returns true because their indices are below 8, even though these byte views require a REX prefix and `validate_single_register` rejects them in x86-32 mode. The search backends and mutator now treat this method as their shared legality filter, so those unencodable registers survive into candidate generation … include the existing Mode32 low-byte/index-4-or-higher restriction here.

**Context:** `src/isa/x86.rs` — `is_available_in` (Mode32 → `index < 8`). The sibling `x86_register_ok` (used by the assembler `can_assemble` prefilter) already rejects these via `!(mode_width == 32 && view == LowByte && index >= 4)`. `is_available_in`'s docstring claims *"True when this GPR can be named as an operand while assembling in `mode`"*, which is inaccurate for SPL/BPL/SIL/DIL in Mode32.

**Why deferred (latent, currently unreachable):** the inconsistency is real, but it is **not a live bug today**:
- The four search register pools are seeded only from `default_x86_registers()` (all Native-view) or `x86_registers_from_target()` (derived from the target's own operands).
- The mode-aware parser `parse_x86_register_for_mode` (`src/parser/x86.rs:198–210`) explicitly rejects SPL/BPL/SIL/DIL in Mode32, so an x86-32 target IR can never contain them → the target-derived pool can never carry them. All other `with_x86_registers(...)` callers are tests passing Native registers.
- Even hypothetically, the assembler's `x86_register_ok` second layer rejects such a register before emission — the worst case would be a wasted candidate proposal, never a wrong result.

PR #681 is a deliberately behavior-preserving refactor of the bare `index < 8` check. Broadening `is_available_in` to the LowByte-aware rule is a legitimate but **separable** hardening that aligns the seam with main's newer `x86_register_ok` model (issue #75); doing it on its own keeps the (unreachable-today) behavior change reviewable on its own merits rather than buried in the refactor.

**Suggested change:** mirror `x86_register_ok`'s register clause in `is_available_in` (reject `LowByte && index >= 4` in Mode32), correct the docstring, and add SIL/DIL Mode32 cases to the `x86_register_availability_by_mode` regression test.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage (both DEFER, high confidence) by Claude Opus 4.8 and Codex GPT-5.6._

pmatos:
## Summary

- route `X86Register::is_available_in` through the shared encoding legality predicate
- reject SPL/BPL/SIL/DIL in x86-32 while preserving legacy low bytes and x86-64 behavior
- document the individual-register contract and add named-register regressions

## Tests

- `cargo test x86_register_availability_by_mode`
- `cargo test register_legality_tracks_rex_availability_per_mode`
- `cargo test --all`
- `./ci_check.sh`

The generic run contract referenced `scripts/full_test.sh`, but that Vow-specific path does not exist in this s11 repository; the repository-defined `./ci_check.sh` gate passed in full.

Closes #684

### DIFF
--- a/src/isa/x86/mod.rs
+++ b/src/isa/x86/mod.rs
@@ -392,19 +392,20 @@ impl X86Register {
         base.canonical().with_view(self.view)
     }
 
-    /// True when this GPR can be named as an operand while assembling in
-    /// `mode`. The extended registers R8..R15 require a REX prefix that does
-    /// not exist in 32-bit mode, so only the eight legacy GPRs are addressable
-    /// there; all sixteen are addressable in 64-bit mode. This is the shared
-    /// source of truth for the x86-32 register-legality constraint that the
-    /// search backends' register pools filter through; the assembler
-    /// `can_assemble` prefilter enforces the same rule via the width-aware
-    /// `x86_register_ok`.
+    /// True when this individual GPR view has an encoding in `mode`.
+    ///
+    /// In 32-bit mode, indices 0..=7 are available, but low-byte views are
+    /// limited to the legacy AL/CL/DL/BL encodings at indices 0..=3. In 64-bit
+    /// mode, all sixteen GPRs and the REX-only low-byte views SPL/BPL/SIL/DIL
+    /// are available. Whole-instruction constraints, such as pairing a legacy
+    /// high-byte register with a REX-requiring operand, remain the encoding
+    /// prefilter's responsibility.
     pub fn is_available_in(self, mode: crate::assembler::x86::X86Mode) -> bool {
-        match mode {
-            crate::assembler::x86::X86Mode::Mode64 => true,
-            crate::assembler::x86::X86Mode::Mode32 => self.index().is_some_and(|i| i < 8),
-        }
+        let mode_width = match mode {
+            crate::assembler::x86::X86Mode::Mode64 => 64,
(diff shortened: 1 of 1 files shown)


pmatos:
Implementation-run note: the generic run contract requested `scripts/full_test.sh`, but that file is absent from s11 and belongs to the Vow workflow described elsewhere in the contract. I did not add an unrelated compatibility script. Instead I ran s11’s canonical `./ci_check.sh`, which builds the required cross-architecture fixtures and passed all policy, formatting, build, Rust test, doc-test, and end-to-end analyzer checks.

PR #780, opened 2026-08-24
refactor(x86): centralize sub-register legality validation

### LINKED ISSUE #680: x86 sub-register views (#675) follow-up: dedup register-legality validation, dead-code cleanup, and dedicated regressions
Deferred from #675: maintainability nits and regression-test gaps flagged in automated review of the x86 sub-register-views PR. Every item was flagged by the reviewer as "not a bug" / "works today"; ADR-0010 already records these test areas as intended follow-ups, so they are tracked here rather than expanding #675.

Original feedback by @claude[bot] on PR #675 (https://github.com/pmatos/s11/pull/675#issuecomment-4990433940).

**Context:** the x86 sub-register-view machinery added in #675 — `src/isa/x86.rs`, `src/assembler/x86.rs`, `src/semantics/*`, `src/parser/x86.rs`, `src/search/*`.

**Why deferred:** scope-creep — maintainability/cosmetic nits plus additive regression tests, none blocking the correctness of a large, coherent, fully-passing feature. The reviewer's concrete-vs-SMT worry was additionally verified manually during the base merge (all SETcc/partial-write concrete and SMT tests pass), lowering urgency without removing the value of permanent cross-checks.

### Design / maintainability
- [ ] **D1 — Consolidate register-legality validation.** `assembler/x86.rs`'s `validate_single_register` / `validate_register_pair` reimplement the "is this register legal in this mode" logic that `isa/x86.rs` already expresses as `x86_register_ok` / `x86_register_pair_ok`. Have the assembler call the `isa::x86` predicates so the rule lives in one place.
- [ ] **D2 — Remove or comment unreachable validation.** `validate_single_register`'s `effective_width > mode_width` check can never trigger; `reg_index`'s `.ok_or_else` arm is unreachable (`X86Register::index()` is total); `isa::x86::canonical()` routes through an `.expect()`-guarded fallible constructor for an infallible op. Simplify or annotate so they don't read as load-bearing validation.
- [ ] **D3 — Trim/confirm unused API.** `X86Register::with_base` (isa/x86.rs:353) has no callers, and `with_view` is only called by `with_base`. (Verified: zero callers.)
- [ ] **D4 — Cross-reference comment.** `main.rs` hardcodes `same_count_code_size_allowed = true`; add a comment linking the ADR-0010 section that justifies it. (Trivial; batch with the rest to avoid a churn-triggering one-line push.)

### Test coverage (additive regressions)
- [ ] **T1 — Direct unit tests** for the register-view helpers/predicates: `canonical()`, `effective_width()`, `is_high_byte()`, `x86_register_ok`, `x86_register_pair_ok` on non-native views (EAX/AX/AL/AH/SPL/R8B, …).
- [ ] **T2 — High-byte + REX rejection** parser regression, e.g. `mov ah, r8b` / `mov dh, spl` (ADR explicitly calls this out).
- [ ] **T3 — Mode32 SPL/BPL/SIL/DIL rejection** in the parser, plus the assembler-side "requires a REX prefix and is not available in x86-32" message.
- [ ] **T4 — Concrete-vs-SMT cross-check** for partial-write / dword zero-extension. The two new equivalence tests `x86_64_smt_models_partial_byte_writes` and `x86_64_smt_models_dword_writes_as_zero_extending` use `.random_tests(0)` and exercise only the SMT path. (Verified.)
- [ ] **T5 — Search-side encodability filter** direct test (e.g. a candidate pool with a high-byte register plus an R8–R15 register).
- [ ] **T6 — `validate_register_pair`** "register widths do not match" rejection test.

Reviewer's suggested priority: **T2**, **T4**, **T1**.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.8 and Codex (GPT-5.6), which both classified this bundle as DEFER (out of scope for #675)._

pmatos:
## Summary

- route assembler single-register and register-pair validation through the shared ISA encodability predicates while preserving actionable diagnostics; this also closes a latent gap where direct x86-32 single-register validation accepted R8-R15 until a later index check
- remove unreachable validation and unused register-view rebinding helpers, make canonicalization directly infallible, and cross-reference ADR-0010 at the frontend policy choice
- add dedicated helper, parser, assembler, concrete-vs-SMT, and search-filter regressions for x86 sub-register views

## Verification

- `cargo fmt --all`
- `cargo test --all` (after building the documented integration fixtures with `./build_tests.sh`)
- `./ci_check.sh`
- all GitHub checks pass, including tests, clippy, coverage/Codecov, CodeQL, title lint, and automated review

The generic implementation-stage prompt named `scripts/full_test.sh`, which does not exist in this s11 repository; `./ci_check.sh` is the repository-defined full-suite gate and passed, including `test_all.sh`.

Closes #680

### DIFF
--- a/src/assembler/x86.rs
+++ b/src/assembler/x86.rs
@@ -17,7 +17,10 @@
 // disfiguring the macro invocations. Allow at module scope.
 #![allow(clippy::useless_conversion)]
 
-use crate::isa::x86::{X86Condition, X86Instruction, X86Register, X86RegisterView};
+use crate::isa::x86::{
+    X86Condition, X86Instruction, X86Register, X86RegisterView, x86_register_ok,
+    x86_register_pair_ok,
+};
 use dynasm::dynasm;
 use dynasmrt::DynasmApi;
 
@@ -56,16 +59,17 @@ impl X86Assembler {
 }
 
 fn reg_index(reg: X86Register) -> Result<u8, String> {
-    reg.index()
-        .ok_or_else(|| format!("register {:?} has no index", reg))
+    // X86Register construction is private and every value stores a valid GPR
+    // index. Keep Result here only so the encoder macros can use one `?`-based
+    // calling convention for index and immediate extraction.
+    Ok(reg
+        .index()
+        .expect("every X86Register has a construction-validated GPR index"))
 }
 
 fn reg_index_32(reg: X86Register) -> Result<u8, String> {
-    let i = reg_index(reg)?;
-    if i >= 8 {
-        return Err(format!("register {:?} not available in x86-32 mode", reg));
-    }
-    Ok(i)
+    validate_single_register(reg, 32)?;
+    reg_index(reg)
 }
 
 fn validate_register_pair(
@@ -75,38 +79,41 @@ fn validate_register_pair(
 ) -> Result<(), String> {
     validate_single_register(lhs, mode_width)?;
     validate_single_register(rhs, mode_width)?;
+
+    if x86_register_pair_ok(lhs, rhs, mode_width) {
+        return Ok(());
+    }
+
+    // The shared predicate above decides legality. The branches below only
+    // classify its remaining pair-specific failure for a useful diagnostic.
     if lhs.effective_width(mode_width) != rhs.effective_width(mode_width) {
         return Err(format!("register widths do not match: {} and {}", lhs, rhs));
     }
-    if (lhs.is_high_byte() || rhs.is_high_byte())
-        && (lhs.index().is_some_and(|index| index >= 4)
-            || rhs.index().is_some_and(|index| index >= 4))
-    {
-        return Err(format!(
-            "high-byte register cannot be encoded when a REX prefix is required: {}, {}",
-            lhs, rhs
-        ));
-    }
-    Ok(())
+    debug_assert!(lhs.is_high_byte() || rhs.is_high_byte());
+    Err(format!(
+        "high-byte register cannot be encoded when a REX prefix is required: {}, {}",
+        lhs, rhs
+    ))
 }
 
 fn validate_single_register(reg: X86Register, mode_width: u32) -> Result<(), String> {
-    if reg.effective_width(mode_width) > mode_width {
-        return Err(format!(
-            "register {} is not available in x86-{}",
-            reg, mode_width
-        ));
+    if x86_register_ok(reg, mode_width) {
+        return Ok(());
     }
-    if mode_width == 32
-        && reg.view() == X86RegisterView::LowByte
-        && reg.index().is_some_and(|index| index >= 4)
-    {
+
+    // The shared predicate is authoritative. A rejected low-byte operand in
+    // Mode32 necessarily belongs to the REX-only SPL..DIL/R8B..R15B family;
+    // distinguish that case solely to retain the actionable encoder message.
+    if mode_width == 32 && reg.view() == X86RegisterView::LowByte {
         return Err(format!(
             "register {} requires a REX prefix and is not available in x86-32",
             reg
         ));
     }
-    Ok(())
+    Err(format!(
+        "register {} is not available in x86-{}",
+        reg, mode_width
+    ))
 }
 
 fn validate_extension_move_registers(
@@ -1419,6 +1426,35 @@ mod tests {
             }])
             .expect_err("AH cannot be encoded in an instruction requiring REX");
         assert!(err.contains("high-byte"), "unexpected error: {err}");
+    }
+
+    #[test]
+    fn x86_32_rejects_rex_only_low_byte_registers_with_actionable_error() {
+        for reg in [
+            X86Register::SPL,
+            X86Register::BPL,
+            X86Register::SIL,
+            X86Register::DIL,
+        ] {
+            let err = X86Assembler::new_32()
+                .assemble_instructions(&[X86Instruction::MovImm { rd: reg, imm: 0 }])
+                .expect_err("x86-32 cannot encode the REX-only low-byte registers");
+            assert_eq!(
+                err,
+                format!("register {reg} requires a REX prefix and is not available in x86-32")
+            );
+        }
+    }
+
+    #[test]
+    fn assembler_rejects_mismatched_register_widths() {
+        let err = X86Assembler::new_64()
+            .assemble_instructions(&[X86Instruction::MovReg {
+                rd: X86Register::RAX,
+                rs: X86Register::EAX,
+            }])
+            .expect_err("x86 register operands must have matching widths");
+        assert_eq!(err, "register widths do not match: rax and eax");
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


pmatos:
Automated code-review pass (`/code-review --fix`) run against this PR. **No correctness bugs found in the diff** — the refactor's core claims verified out:

- `x86_register_pair_ok`'s high-byte clause negates to exactly the old assembler conjunction, so the new `debug_assert!` in `validate_register_pair` cannot fire and the unconditional fall-through error is correct in release builds too.
- `with_view`/`with_base` have no remaining callers repo-wide; the deletion is safe, including the `HighByte` index>=4 guard they carried, since no public constructor can build that value.
- The claimed latent gap is real: the old `validate_single_register` width check was dead for every view, so x86-32 single-register validation did accept R8..R15 until `reg_index_32`'s later index check.

Two findings were fixed in e6e4545:

1. **`src/isa/x86/mod.rs` — `is_available_in` duplicated and weakened `x86_register_ok`.** Its doc comment claimed to be the shared source of truth but only tested `index < 8`, so it admitted SPL/BPL/SIL/DIL in Mode32 where `x86_register_ok(reg, 32)` rejects them. It now delegates. Mode64 is provably unchanged (`x86_register_ok(reg, 64)` reduces to the always-true `index < 16`).

2. **`src/semantics/equivalence.rs` — raising `random_tests` 0→16** let the concrete fast path refute the negative assertion in `x86_64_concrete_and_smt_model_partial_byte_writes` before SMT was reached, so that assertion no longer covered the SMT partial-write model. The pair is now asserted both concretely and under `random_tests(0)` (the established skip-the-fast-path idiom already used at lines 1744/1911/1928). Deliberately *not* mirrored into `..._dword_writes_as_zero_extending`: it has only a positive assertion and the fast path returns `None` there, so SMT still runs.

**Judgement call worth flagging:** fix (1) touches a function *outside* this PR's hunks — the diff removed `with_view`/`with_base` immediately above it but never touched `is_available_in` itself. Its only new effect is excluding SPL/BPL/SIL/DIL from Mode32 search pools. All four consumers (`X86Mutator::new`, the `X86_32` symbolic and enumerative `registers_from_config`, `x86_random_sequence`) and both pool sources were traced: `default_x86_registers` is Native-only, and the Capstone Mode32 path that feeds `registers_from_target` cannot produce SPL. It is a no-op for every current caller and can only ever remove candidates the assembler would reject anyway. Revert it if you'd rather keep the PR strictly to its own hunks.

Two things deliberately left alone: `reg_index` returning an always-`Ok` `Result` (the uniform `?` convention for the encoder macros is documented inline; de-`Result`ing touches ~30 call sites for no behavioral gain), and the redundant double validation in the `encode_32` arms (harmless and cheap; restructuring would churn the diff).

Gate: `./ci_check.sh` green end to end, verified independently after the fixes — `cargo fmt --check`, clippy clean, 1706 + 101 + 67 tests pass.

pmatos:
Rebased onto `origin/main` (now includes #779 `fix(x86): reject REX-only byte registers in 32-bit mode` and #782 `perf(auto): inspect Capstone detail once per instruction`) and force-pushed with `--force-with-lease`. `mergeable` is back to `MERGEABLE`.

## Conflicts and how they were resolved

Both conflicts were in `src/isa/x86/mod.rs`, and both came from **#779 landing an independent, functionally identical fix to `is_available_in`** while this branch was open. Nothing here required a product decision; each resolution preserves both sides.

1. **`10f63e1` — dead-code deletion adjacent to a rewritten doc comment.** This branch deletes `with_view`/`with_base`; #779 rewrote the `is_available_in` doc comment immediately below them. Took main's rewritten doc comment *and* applied this branch's deletion. Re-confirmed `with_view`/`with_base` still have no callers anywhere in `src/` after #779 and #782.

2. **`e6e4545` — duplicate delegation to `x86_register_ok`.** Both sides make `is_available_in` delegate to `x86_register_ok(self, mode_width)`; the bodies are identical modulo path spelling. Kept the unqualified `x86_register_ok`, which this branch re-exports at `src/isa/x86/mod.rs:21` for the assembler refactor.
   - **Doc comment:** merged rather than picked. Main's wording is the more precise statement of the rule (indices `0..=7`, low-byte views limited to `0..=3`, whole-instruction constraints left to the prefilter); this branch's added the *why* (mode-typed spelling of `x86_register_ok`, so search pools filter through exactly what the assembler enforces). Both are now in the one comment.
   - **`x86_register_availability_by_mode` test:** unioned both register lists instead of taking a side. #779's version covers `AL`/`CL`/`DL`/`BL` and `SPL`/`BPL`/`SIL`/`DIL`; this branch's covers `EAX`/`AX`/`AL`/`AH` and `R8D`/`R8B`. The union exercises the dword, word, legacy-low-byte and legacy-high-byte views on both the available and rejected sides — strictly more coverage than either side alone. `AH` is a `HighByte` view at index 0, so `x86_register_ok`'s `LowByte && index >= 4` clause does not apply and it stays available in Mode32; verified, and the test passes.

`e6e4545`'s commit message was amended so it no longer claims a code change that main already made — the delegation now arrives via #779, and what the commit still contributes is the merged rationale, the wider availability test, and the `equivalence.rs` `random_tests(0)` fix. Its diff is otherwise unchanged.

## Judgment call: an uncommitted local refinement pass was deliberately not folded in

The reused workspace contained an **uncommitted** further refactor (~200 diff lines across `assembler/x86.rs`, `isa/x86/{encoding,mod,stochastic}.rs`, `parser/x86.rs`, `search/symbolic/backend.rs`), left behind by the previous attempt after `e6e4545` was pushed. It added `X86Mode::width()`/`from_width()` helpers, routed `parse_x86_register_for_mode` and `validate_same_register_width` through the shared predicates, and refreshed several comments.

I set it aside rather than including it, for three reasons:

- This run's mandate is conflict resolution; folding an unreviewed refactor into an already-reviewed PR widens scope without review.
- Its `validate_same_register_width` hunk is a behavior change, not a cleanup: it replaces the trailing `Ok(())` with an unconditional `Err(...)` guarded only by a `debug_assert!`. The equivalent assembler-side pattern got explicit reviewer scrutiny in this PR; the parser-side one has had none, and the parser's precondition story differs (the `None`/64-bit path).
- It was authored against the pre-#779 `is_available_in`, which #779 has since rewritten. It re-applied without textual conflict, which is exactly the case where a clean apply does not imply it is still semantically intended.

It is preserved in the workspace under `.clawpatch/s11-680-artifacts/` (`uncommitted-refinement.patch` pre-rebase, `uncommitted-refinement-rebased.patch` post-rebase; also copied to `target/s11-680-artifacts/`, which `cargo clean` would wipe) if you want it as a follow-up PR. Nothing is lost by not taking it now.

## Gate

`./ci_check.sh` green end to end on the rebased branch: repository CI-policy checks, mutation-wrapper regressions, `cargo fmt -- --check`, build, `build_tests.sh`, `cargo test` (1706 + 102 + 67 passing), and `test_all.sh`. `cargo clippy --all -- -D warnings` (the PR's documented invocation) is clean.

One unrelated note: `cargo clippy --all --all-targets -D warnings` reports two `clone_on_copy` errors in `src/x86_window_reassembly.rs:245,280`. These are pre-existing on `main` — the file is untouched by this branch and `X86Instruction` has derived `Copy` on both sides — and CI's clippy job runs with `continue-on-error: true` and no `-D warnings`, so it does not gate this PR. Flagging it as a separate cleanup, not fixing it here.

PR #782, opened 2026-08-24
perf(auto): inspect Capstone detail once per instruction

### LINKED ISSUE #686: Cache Capstone InsnDetail across --auto window-finder phases to avoid double decode
Deferred from #683: the `--auto` window-finder decodes `InsnDetail` twice per instruction.

Original feedback by the `claude[bot]` reviewer on PR #683 (https://github.com/pmatos/s11/pull/683):

> Phase 1 calls `cs.insn_detail(instruction)` on *every* instruction in every section just to feed `capstone_detail_direct_branch_targets`. Phase 2 calls `cs.insn_detail` again on the same instructions for the call/RIP-relative checks. That's a full second detail-decode pass over the whole binary's instruction stream. Given the ADR explicitly frames this driver as running over binaries the size of `/bin/ls`, it'd be worth caching the `InsnDetail` (or at least the extracted target list) from phase 1 in `decoded_sections` and reusing it in phase 2 rather than recomputing. Not correctness-affecting, just avoidable work.

**Context:** `src/main.rs`, `find_candidate_windows_with_backend` — phase 1 (~line 1175) calls `cs.insn_detail` per instruction to collect branch targets; phase 2 (~line 1232) calls it again for the call / RIP-relative checks.

**Why deferred:** diminishing-returns — the discovery path is still `#[cfg_attr(not(test), allow(dead_code))]` (not yet CLI-wired), runs once per binary at discovery time, and caching `InsnDetail` across the two phases fights its borrow lifetime (the extracted per-instruction target/flag lists would need to be stored instead). Worth doing when the driver (#620) is wired and the path is on a hot binary-scale loop, not before.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.8 and Codex._

pmatos:
## Summary

- reduce each borrowed Capstone instruction detail to owned branch, call, and RIP-relative facts during phase 1
- cache classified `WindowInstruction` descriptors for phase 2 while preserving binary-wide target collection before planning
- add an ELF/Capstone regression proving detail inspection runs once per decoded instruction while call and RIP-relative exclusions remain intact

## Validation

- `cargo fmt --all`
- `cargo test --all`
- `./ci_check.sh`

Closes #686

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1070,16 +1070,75 @@ fn find_candidate_windows_with_backend<B: ElfOptimizationBackend>(
     backend: B,
     patcher: &ElfPatcher,
 ) -> Result<Vec<SectionCandidateWindows>, Box<dyn std::error::Error>> {
+    find_candidate_windows_with_detail_provider(
+        backend,
+        patcher,
+        inspect_capstone_instruction_detail,
+    )
+}
+
+/// Owned facts reduced from one borrowed Capstone detail inspection.
+///
+/// Keeping only planning inputs lets the two finder phases share the result
+/// without extending `InsnDetail`'s borrow across section processing.
+#[derive(Debug)]
+struct CapstoneInstructionFacts {
+    direct_branch_targets: Vec<u64>,
+    is_call: bool,
+    has_rip_relative_memory: bool,
+}
+
+fn inspect_capstone_instruction_detail(
+    cs: &Capstone,
+    instruction: &capstone::Insn<'_>,
+    section_name: &str,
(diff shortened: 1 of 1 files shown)


pmatos:
Autonomous-run note: the implementation prompt referenced `scripts/full_test.sh`, but that Vow-specific script does not exist in this s11 repository. I used the repository-mandated `./ci_check.sh` together with `cargo fmt --all`, zero-warning Clippy, and `cargo test --all`; all pass. The prompt also named `agent-ready`, while issue #686 actually carried `ready-for-agent`, so I removed the existing readiness label to prevent rescheduling.

Issue #1, opened 2025-05-26
Feature: Turn flags into commands like in git


pmatos:
Flags `--disasm` and `--opt` should really be command like:
  * s11 disasm <disasm options>...
  * s11 opt <opt options>...
  * s11
    * s11 without a valid command should print help and exit.

each mode followed up by -h or --help should bring the help for that mode. This follows the structure of programs like git.

pmatos:
@claude can you implement this feature?

pmatos:
@claude can you create a PR with the changes you committed to f31fbb7 ?

PR #784, opened 2026-08-24
test(auto): cover AArch64 TBZ/TBNZ target over-collection

### LINKED ISSUE #688: Add a test exercising AArch64 tbz/tbnz bit-position over-collection in direct-target scan
Deferred from #683: a documented invariant of `capstone_detail_direct_branch_targets` is not exercised by a test.

Original feedback by the `claude[bot]` reviewer on PR #683 (https://github.com/pmatos/s11/pull/683#issuecomment-5004280362):

> The doc comment explicitly notes AArch64 `tbz`/`tbnz`'s bit-position immediate is "harmlessly over-collected" as a target, but that specific claim isn't exercised by a test — a regression there wouldn't be caught.

**Context:** `src/main.rs`, `capstone_detail_direct_branch_targets`. The doc comment reasons that a `tbz x0, #5, label` yields both the bit-position immediate (5) and the branch target, and that the extra value is harmless (an extra split at worst, never an unsound admit). No test pins this two-immediate behavior.

**Why deferred:** diminishing-returns / test-coverage — the reviewer flagged it as non-blocking ("worth a follow-up test if convenient"). A small disasm-based test asserting the collected set for `tbz`/`tbnz` would guard the documented claim against a future filter change (e.g. one that deduplicates or drops non-target immediates).

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.8 and Codex._

pmatos:
Adds a disassembly-backed regression for both AArch64 test-branch forms.

- pins `tbz` bit 5 and its absolute target in operand order
- pins `tbnz` bit 40 and its absolute target in operand order
- verifies the documented conservative over-collection behavior without production changes

Closes #688

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3881,6 +3881,49 @@ mod cli_helper_tests {
     }
 
     #[test]
+    fn direct_branch_targets_overcollect_aarch64_tbz_tbnz_bit_positions() {
+        let cs = aarch64_test_capstone();
+        for (mnemonic, instruction, bit) in [
+            (
+                "tbz",
+                Instruction::Tbz {
+                    rt: Register::X0,
+                    bit: 5,
+                    target: s11::ir::LabelId(0x1100),
+                },
+                5,
+            ),
+            (
+                "tbnz",
+                Instruction::Tbnz {
+                    rt: Register::X0,
+                    bit: 40,
+                    target: s11::ir::LabelId(0x1100),
+                },
+                40,
+            ),
+        ] {
+            let bytes = assemble_aarch64_test_bytes(&[instruction]);
+            let disassembly = cs
+                .disasm_all(&bytes, 0x1000)
+                .unwrap_or_else(|_| panic!("{mnemonic} fixture should disassemble"));
+            let instruction = disassembly
+                .iter()
+                .next()
+                .unwrap_or_else(|| panic!("fixture should contain {mnemonic}"));
+            let detail = cs
+                .insn_detail(instruction)
+                .unwrap_or_else(|_| panic!("{mnemonic} detail should be available"));
+
+            assert_eq!(
+                capstone_detail_direct_branch_targets(&detail),
+                vec![bit, 0x1100],
+                "{mnemonic} bit position and absolute branch target must both be collected"
+            );
+        }
+    }
+
+    #[test]
     fn candidate_windows_split_at_interior_direct_branch_target() {
         // add rax,1 (0x1000); add rax,1 (0x1004); add rax,1 (0x1008);
         // jne 0x1004 (0x100c). Without the target split this is one window


pmatos:
Run-contract note: the generic implementation instructions referenced `scripts/full_test.sh`, but that Vow-specific script does not exist in this s11 repository. I followed s11’s repository-local gate instead: strict Clippy, `./ci_check.sh` (which builds the required ELF fixtures and runs the complete suite), and a subsequent successful `cargo test --all`. The alternative—adding an unrelated gate script—would have expanded this test-only issue beyond scope.

pmatos:
## Automated code-review pass (post-merge)

Ran `/code-review --fix` at effort `xhigh` against this diff. **This PR squash-merged as `bce1762` while the review was in flight**, so the resulting fixes could not land here.

Per this run's contract I did **not** open a second PR. The fixes are committed, rebased onto current `main`, and pushed to the original branch as a single clean commit:

- Branch: `sym/s11/688-add-a-test-exercising-aarch64-tbz-tbnz-bit-position-over-collection-in-direct-target-scan`
- Commit: `cdc27f4` — `test(auto): tighten TBZ/TBNZ over-collection fixture`
- Delta vs `main`: `src/main.rs` only, +24/-2, test-only

Open a PR from that branch to land it.

### Applied fixes (5)

1. **Mnemonic assertion** — the table threaded `mnemonic` through the loop but used it only in message strings, so nothing verified what Capstone actually decoded. Added `assert_eq!(decoded.mnemonic(), Some(mnemonic))`. Proved it has teeth: swapping the near-identical `Tbz`/`Tbnz` dynasm arms in `src/assembler/mod.rs:2177-2196` passed the old test and fails the new one (`left: Some("tbnz")`, `right: Some("tbz")`). Assembler restored afterwards.
2. **`assert_eq!(disassembly.len(), 1)`** — the test read `disassembly.iter().next()` without pinning the decode count, so a multi-instruction expansion for `Tbz` would have silently covered only half the fixture.
3. **Bit-0 table row** — the table covered only interior bit values (5, 40), skipping the boundary value most likely to collide with a real address.
4. **Renamed the shadowed `instruction`** to `decoded` — the loop pattern's `s11::ir::Instruction` and `capstone::Insn` shared one name inside a 20-line body.
5. **Rationale comment** — the assertion message frames the over-collection as a hard requirement; the comment marks it conservative-by-design so a future narrowing of the scan doesn't read as a regression.

Verified: `./ci_check.sh` green pre-rebase; `cargo test --bin s11` 103 passed / 0 failed and `cargo fmt -- --check` clean post-rebase (separate commands per CLAUDE.md).

### Report-only (deferred — all production changes in a PR that advertised none)

- **`src/main.rs:1359` — the doc rationale is falsifiable.** "a 0..=63 bit index never coincides with a real in-section code address" does not hold: `get_text_sections` has no `e_type` filter, so an AArch64 ET_REL object whose `.text` has `sh_addr = 0` is accepted by `ElfPatcher::new` (`src/elf_patcher/mod.rs:129`) and reports `virtual_addr = 0`. A `tbnz x0, #40, ...` in that object inserts `40` into `branch_targets`, and `plan_candidate_windows` then splits at the real instruction address `0x28`. Note this interacts with fix #3 above: the new row pins `bit = 0`, exactly the colliding value. The test is still correct — it asserts that over-collection happens, which it does — but its comment points at a rationale with this known soft spot.
- **`src/main.rs:1360` — "extra split, never an unsound admit" is stronger than what holds.** An extra split can flip a *refused* window into an *admitted* one: if the spurious target coincides with an indirect target, that target becomes the first instruction of the second window, and `refuse_windows_with_interior_targets` (`src/candidate_windows.rs:147`) only rejects `start < target < end`. Sound per ADR-0009 (a window may begin at a target), but the prose asserts more than that.
- **Altitude: the over-collection is avoidable at the source.** On AArch64 the branch target is always the *last* immediate of a direct branch (`b`/`bl`/`b.cond`: one imm; `cbz`/`cbnz`: reg + imm; `tbz`/`tbnz`: reg + imm(bit) + imm(target)). Taking the last immediate removes the hazard rather than documenting it. This argues against an ADR-0009 design decision — author's call.
- **Boilerplate: 9th copy** of the disassemble → first-insn → `insn_detail` → assert-targets block (`src/main.rs:3773, 3782, 3792, 3805, 3827, 3848, 3875, 3917`). A helper returning `Vec<u64>` would escape the `InsnDetail` borrow chain, but doing so structurally forecloses the mnemonic assertion in fix #1, and the other 8 sites are outside this diff.

### One correction outside the diff

The PR body says to merge with `gh pr merge --merge` "per CLAUDE.md". No such rule exists — the repo `CLAUDE.md` has no merge-strategy guidance at all. (This PR was in fact squash-merged.) Worth dropping the citation before anyone relies on it.

PR #792, opened 2026-08-25
refactor(aarch64): clarify move-wide alias bindings

### LINKED ISSUE #441: Rename shadowed `imm` binding in normalize_mov_wide_alias for readability
Deferred from #414: a readability nit on pre-existing code in `normalize_mov_wide_alias` (src/main.rs).

Original feedback from the automated `claude[bot]` review on PR #414 (https://github.com/pmatos/s11/pull/414):

> **Variable shadowing in `normalize_mov_wide_alias`** (minor readability)
>
> The outer `imm: i64` (from `parse_immediate`) and the inner `(imm, shift)` pairs (from the encoding functions, where `imm: u16`) share the same binding name. This is valid Rust but a reader has to look up both types. Renaming the inner bindings to `(enc_imm, enc_shift)` would remove any potential for confusion.

**Context:** `normalize_mov_wide_alias` in `src/main.rs` — outer `let imm = parser::parse_immediate(...)` (`i64`) is shadowed by the inner `if let Some((imm, shift)) = move_wide_movz_encoding(value)` (`u16`). The inner `imm` lives only inside an immediately-returning block, so there is no behavioral risk.

**Why deferred:** diminishing-returns — pure readability/style on pre-existing code with no behavioral impact. The two independent reviewers split (DEFER vs REJECT) on whether it is worth tracking at all; filing this so it is not silently dropped, but it is low priority. Note the same `claude-review` bot has issued contradictory guidance across runs on this file (it called the `imm != 0` guard both "dead code" and "necessary"), so treat its style nits with appropriate skepticism.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage._

pmatos:
## Summary

- rename the encoded MOVZ and MOVN immediate/shift tuple bindings to `enc_imm` and `enc_shift`
- keep parsed source immediates visibly distinct from encoded move-wide fields without changing normalization behavior

## Testing

- `cargo test --lib capstone_bridge::tests::convert_capstone_op_normalizes_mov_wide_aliases` (before and after)
- `cargo fmt --all`
- `just build-tests && cargo test --all`
- `./ci_check.sh`

The Vow-template command `scripts/full_test.sh` is not present in this s11 repository; the project-defined comprehensive `./ci_check.sh` gate passed instead.

Closes #441

### DIFF
--- a/src/capstone_bridge.rs
+++ b/src/capstone_bridge.rs
@@ -84,11 +84,11 @@ fn normalize_mov_wide_alias(op_str: &str) -> Result<Option<String>, String> {
     }
 
     let value = imm as u64;
-    if let Some((imm, shift)) = move_wide_movz_encoding(value) {
-        return Ok(Some(format_move_wide("movz", rd, imm, shift)));
+    if let Some((enc_imm, enc_shift)) = move_wide_movz_encoding(value) {
+        return Ok(Some(format_move_wide("movz", rd, enc_imm, enc_shift)));
     }
-    if let Some((imm, shift)) = move_wide_movn_encoding(value) {
-        return Ok(Some(format_move_wide("movn", rd, imm, shift)));
+    if let Some((enc_imm, enc_shift)) = move_wide_movn_encoding(value) {
+        return Ok(Some(format_move_wide("movn", rd, enc_imm, enc_shift)));
     }
 
     Ok(None)


pmatos:
Implementation judgment: this issue is an identifier-only refactor, so no behavioral test can be made red by the requested change without asserting Rust source spelling. I used the existing `convert_capstone_op_normalizes_mov_wide_aliases` public-seam regression as a green-before/green-after characterization test; the alternative source-text assertion would be brittle and implementation-coupled. The Vow-template `scripts/full_test.sh` command is also absent from s11, so the repository-defined `./ci_check.sh` comprehensive gate was used and passed.

PR #795, opened 2026-08-25
docs(search): clarify candidate-pool cache intent

### LINKED ISSUE #461: Add intent comments to enumerative candidate-pool cache (reset preservation + with_config discard)
Deferred from #447: two low-value documentation nits on the enumerative candidate-pool cache.

Original feedback by `claude[bot]` (automated review) on PR #447 (https://github.com/pmatos/s11/pull/447):

> **1.** `reset()` silently drops the pool-preservation contract — add a one-line comment noting `candidate_pool` is intentionally *not* cleared across resets.
>
> **4.** `let _ = …` in `with_config` reads like an accident — add a short comment clarifying the discard is a deliberate cache-populating side-effect call.

**Context:** `src/search/enumerative/search.rs` — `reset()` and `with_config()`.

**Why deferred:** diminishing-returns. Both are comment-only additions on code whose intent is already pinned elsewhere — the `reset_preserves_candidate_pool` test documents and enforces the preservation contract, and `with_config`'s existing doc comment plus the `_for_config` method name already convey the eager-population intent. The project follows a "comment sparingly, only comment complex code" convention, so adding these now is churn without proportional value. Best bundled into the next change that touches this file.

_Filed automatically by `pm-autofix-pr` after dual-evaluator triage by Claude Opus 4.8 and Codex (gpt-5.5)._

pmatos:
Clarify that `with_config` deliberately discards the returned slice after eagerly populating the enumerative candidate-pool cache.

Document that `reset` intentionally preserves the generated candidate pool for reuse while clearing per-run statistics. This is a comment-only change; the existing focused regression tests continue to enforce both behaviors.

Closes #461

### DIFF
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -808,6 +808,7 @@ where
     }
 
     fn reset(&mut self) {
+        // Preserve the candidate pool so its generated instructions remain reusable across resets.
         self.statistics = SearchStatistics::new(Algorithm::Enumerative);
     }
 }
--- a/src/search/enumerative/search.rs
+++ b/src/search/enumerative/search.rs
@@ -387,6 +387,7 @@ where
     /// generated up front.
     pub fn with_config(config: &SearchConfig) -> Self {
         let mut search = Self::new();
+        // Populate the cache eagerly; the returned slice is intentionally discarded.
         let _ = search.candidate_pool_for_config(config);
         search
     }


pmatos:
Ran `/code-review --fix` (xhigh) against this diff. No correctness bugs; three comment-quality findings, all applied in 3a5b389.

- `reset()` — the original comment said the pool is preserved but not *why it is load-bearing*: `search()` calls `reset()` as its first statement, so clearing `candidate_pool` would defeat the cache on every search, not just on explicit resets. The comment now says that, and also covers `private_pool` (the cached rayon `ThreadPool`), which survives `reset()` for the same reason and was previously undocumented.
- `with_config()` — the original comment restated the code (`let _ =` → "intentionally discarded"). The review's rewrite explained warm-cache semantics but dropped the discard rationale that issue #461 explicitly asked for, so I restored it. It also read as if `search()` unconditionally re-derives the pool; `candidate_pool_for_config` keys on `(registers, immediates)` and only regenerates on a mismatch, which the final wording reflects.

Two review suggestions were skipped deliberately: rewriting `let _ = ...` to a bare statement (pure churn — the return type is not `#[must_use]`, so neither form warns), and a general "comment sparingly" objection (unactionable against a PR whose whole purpose is adding these comments).

Out of scope but worth flagging: `EnumerativeSearch::with_config` has no production callers — `src/main.rs:2157` and `src/bench_support.rs:276` both use `::new()`, and the only call sites are two unit tests. It is `pub`, so `dead_code` never fires. Pre-existing and untouched here.

`./ci_check.sh` passes on the final state.

